PR #4, opened 2017-05-02
Skip the NoPIA-related tests on Mono runtime.

KirillOsenkov:
Adding a SkipOnMono property on MefFact. If it's set, and we're running on the Mono runtime, skip the test.

Mark the NoPIA-related tests with a Trait.


KirillOsenkov:
I've confirmed that the tests are not excluded on Windows and are skipped on Mono:
image

KirillOsenkov:
I'm thinking on those Traits I should swap names and values, so that it's:
[Trait("SkipOnMono", "NoPIA")

Because I'm also adding in a new PR:
[Trait("SkipOnMono", "WeakReference")]
and 
[Trait("SkipOnMono", "AppDomains")]

src/tests/Microsoft.VisualStudio.Composition.Tests/XunitExtensions/MefFactDiscoverer.cs
```diff
@@ -70,6 +70,11 @@ public MefFactTestCase(IMessageSink diagnosticMessageSink, TestMethodDisplay def
                 this.compositionVersions = factAttribute.CompositionVersions;
                 this.noCompatGoal = factAttribute.NoCompatGoal;
                 this.invalidConfiguration = factAttribute.InvalidConfiguration;
+
+                if (factAttribute.SkipOnMono && Type.GetType("Mono.Runtime") != null)
```
AArnott: Does this code have visibility into the Traits on the test such that we can use that skip instead of a new property on the attribute?

src/tests/Microsoft.VisualStudio.Composition.Tests/EmbeddedableTypesMixedTests.cs
```diff
@@ -20,7 +20,8 @@ namespace Microsoft.VisualStudio.Composition.Tests
     /// </summary>
     public class EmbeddedableTypesMixedTests
     {
-        [MefFact(CompositionEngines.V1Compat, "Microsoft.VisualStudio.Composition.NonEmbeddingTypeReceiver", typeof(EmbeddableTypesTests.PartThatExportsIVsProjectReference))]
+        [MefFact(CompositionEngines.V1Compat, "Microsoft.VisualStudio.Composition.NonEmbeddingTypeReceiver", typeof(EmbeddableTypesTests.PartThatExportsIVsProjectReference), SkipOnMono = true)]
+        [Trait("NoPIA", "SkipOnMono")]
```
AArnott: I'd prefer the "NoPIA" trait have a value of "true" rather than "SkipOnMono" since the value doesn't mean anything -- and the way you have it here, it looks like the value is somehow significant.

What I was hoping for was two traits:
```csharp
[Trait("NoPIA", "true")]
[Trait("SkipOnMono", "true")]
```

And then have the test discovery system skip on mono based on the second of these two.

AArnott:
Also: consider adding mono testing to appveyor.yml so that we don't regress it once we achieve a clean run on mono.

src/tests/Microsoft.VisualStudio.Composition.Tests/XunitExtensions/MefFactDiscoverer.cs
KirillOsenkov: I haven't found such a way. As far as I can tell it can only see its own attribute, not the other attributes. If you find a way, let me know.

src/tests/Microsoft.VisualStudio.Composition.Tests/EmbeddedableTypesMixedTests.cs
KirillOsenkov: I can certainly change traits to look like that. However I haven't found a way to drive the skipping logic off of a trait.

Shall I still change the traits like you suggest above? I'm assuming I'll also be adding:

```
[Trait("AppDomains", "true")]
and
[Trait("WeakReference", "true")]
```
as other causes of test failures I've identified so far.

KirillOsenkov:
Yes, we can certainly do appveyor after we have a clean run. For now it's too crashy to even try.

PR #6, opened 2017-05-08
Work toward compiling on .netstandard (1.6)

AArnott:
This fixes most (all the easy) compile errors when compiling on netstandard1.6. It doesn't actually enable such a compilation so that it can still be merged with master.

The changes consist of:

1. Use `TypeInfo` instead of `Type` where required.
2. Reference .netstandard compatible versions of NuGet packages.
3. Remove references to .NET MEF, which isn't available on .netstandard.
4. Replace use of `Thread` with the managed thread ID for purposes of detecting thread equivalence.

### DIFF
--- a/src/Microsoft.VisualStudio.Composition/Reflection/MemberRef.cs
+++ b/src/Microsoft.VisualStudio.Composition/Reflection/MemberRef.cs
@@ -63,9 +63,9 @@ namespace Microsoft.VisualStudio.Composition.Reflection
                     this.Property = new PropertyRef((PropertyInfo)member, resolver);
                     break;
                 default:
-                    if (member is TypeInfo type)
+                    if (member is TypeInfo typeInfo)
                     {
-                        this.Type = TypeRef.Get(type.AsType(), resolver);
+                        this.Type = TypeRef.Get(typeInfo.AsType(), resolver);
                     }
                     else
                     {
--- a/src/Microsoft.VisualStudio.Composition/Reflection/MemberRef.cs
+++ b/src/Microsoft.VisualStudio.Composition/Reflection/MemberRef.cs
@@ -63,9 +63,9 @@ namespace Microsoft.VisualStudio.Composition.Reflection
                     this.Property = new PropertyRef((PropertyInfo)member, resolver);
                     break;
                 default:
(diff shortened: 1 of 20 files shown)


src/Microsoft.VisualStudio.Composition/Reflection/MemberRef.cs
```diff
@@ -63,9 +63,9 @@ public MemberRef(MemberInfo member, Resolver resolver)
                     this.Property = new PropertyRef((PropertyInfo)member, resolver);
                     break;
                 default:
-                    if (member is Type)
+                    if (member is TypeInfo type)
```
AlexEyler: nit: consider "typeInfo"

src/Microsoft.VisualStudio.Composition/ReflectionHelpers.cs
```diff
@@ -207,10 +207,10 @@ internal static bool IsStatic(this MemberInfo exportingMember)
 
         internal static Type GetMemberType(MemberInfo fieldOrPropertyOrType)
         {
-            var type = fieldOrPropertyOrType as Type;
+            var type = fieldOrPropertyOrType as TypeInfo;
```
AlexEyler: nit: here as well (consider typeInfo)

AArnott: fixed

src/Microsoft.VisualStudio.Composition/Reflection/MemberRef.cs
AArnott: fixed

AArnott:
The remaining work for .netstandard1.6 is harder. It involves changing the way we cache and invoke members using handles and metadata tokens. 
As we look at making the cache more resilient against MEF catalog assembly changes, one option would take us significantly further away from reaching netstandard. But as I think we have a couple approaches we could take, I think reaching netstandard first, then fixing the cache issue, may make it more likely we can achieve both desirable ends instead of just one.

AlexEyler:
Ah right - I forgot the metadata tokens weren't supported in .netstandard as well.

PR #8, opened 2017-05-19
Better failure messages when assertions fail in AppDomain tests

### LINKED ISSUE #7: Unit-tests that run in another AppDomain should use serializable asserts
When all unit-tests are passing, everything is OK. However imagine a test breaks (actually true on Mono), instead of getting a readable assert you get:

```
Type 'Xunit.Sdk.FalseException' in assembly 'xunit.assert, Version=2.0.0.2929, Culture=neutral, PublicKeyToken=8d05b1bb7a6fdb6c' is not marked as serializable.
```

This is because xUnit's assert exceptions aren't serializable (and it's by design unfortunately), so you don't get a good message about the failure.

To simulate this, insert an `Assert.False(true);` at this line for example:
https://github.com/Microsoft/vs-mef/blob/master/src/tests/Microsoft.VisualStudio.Composition.Tests/AssembliesLazyLoadedTests.cs#L296

Then run the `ComposableAssembliesLazyLoadedByLazyImport` test.

Maybe it's worth creating a custom serializable AssertException and have an Assert() helper method in AppDomain tests that throws that and preserves a good message when serialized across appdomains?

KirillOsenkov:
If xUnit asserts fail in another appdomain, instead of the legible error we get
Type 'Xunit.Sdk.FalseException' in assembly 'xunit.assert, Version=2.0.0.2929, Culture=neutral, PublicKeyToken=8d05b1bb7a6fdb6c' is not marked as serializable.

Instead, we have our own serializable exception that can travel across appdomains and preserve a better error message in case any of these tests fail.

Fixes #7

### DIFF
--- a/src/tests/Microsoft.VisualStudio.Composition.Tests/AssembliesLazyLoadedTests.cs
+++ b/src/tests/Microsoft.VisualStudio.Composition.Tests/AssembliesLazyLoadedTests.cs
@@ -15,6 +15,7 @@ namespace Microsoft.VisualStudio.Composition.Tests
     using Microsoft.VisualStudio.Composition.AppDomainTests;
     using Microsoft.VisualStudio.Composition.AppDomainTests2;
     using Xunit;
+    using static Microsoft.VisualStudio.Composition.Tests.AssertEx;
 
     [Trait("Efficiency", "LazyLoad")]
     public abstract class AssembliesLazyLoadedTests : IDisposable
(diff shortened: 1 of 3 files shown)


src/tests/Microsoft.VisualStudio.Composition.Tests/XunitExtensions/AssertFailedException.cs
```diff
@@ -0,0 +1,28 @@
+﻿namespace Microsoft.VisualStudio.Composition.Tests
+{
+    using System;
+    using System.Runtime.Serialization;
+
+    [Serializable]
+    internal class AssertFailedException : Exception
+    {
+        public AssertFailedException()
```
AArnott: nit: can you use the `internal` modifier on members that are actually internal (by virtue of their internal container)? That way, my public API cost alarm doesn't go off during code reviews. :)

src/tests/Microsoft.VisualStudio.Composition.Tests/AssembliesLazyLoadedTests.cs
```diff
@@ -270,87 +270,130 @@ internal void Initialize(Type cacheManagerType, Stream cachedComposition, Stream
                 this.container = containerFactory.CreateExportProvider();
             }
 
+            /// <summary>
+            /// Since the exceptions thrown by xUnit asserts are not serializable we lose information
+            /// in case tests fail in another appdomain. These four helpers throw a serializable exception
+            /// that results in a more information exception in the error log.
+            /// </summary>
+            internal void AssertFalse(bool condition, [CallerFilePath] string filePath = null, [CallerMemberName] string memberName = null, [CallerLineNumber] int lineNumber = 0)
```
AArnott: Can these methods be made static?
Can we move them to another class? (`AssertEx`, perhaps)

src/tests/Microsoft.VisualStudio.Composition.Tests/XunitExtensions/AssertFailedException.cs
KirillOsenkov: Fixed. But hopefully you're aware that this practice makes it harder to switch to public in the future (instead of fixing the modifier on the type you also need to fix on all members). I'm guessing this is what you want though.

KirillOsenkov:
Honestly I don't think I like wrap and rethrow, but I can't articulate why. I think we only need it in this one place so I think as a tactical fix to unblock my Mono testing this should be good enough?

KirillOsenkov:
Let me know if you think it's OK for this to go in.

src/tests/Microsoft.VisualStudio.Composition.Tests/XunitExtensions/AssertFailedException.cs
AArnott: Yes: public APIs have a very real long term cost. I want to make sure I only pay for it for the individual members that need it. :)
Of course it's much less of a problem (if one at all) in test assemblies.

AArnott:
> Honestly I don't think I like wrap and rethrow, but I can't articulate why.

I can articulate why I prefer the wrap and rethrow approach: The specialness of these tests are at the AppDomain boundary, so I'd rather that be where we centralize the specialized code. Otherwise it's going to be hard to remember why these tests are "special" and to keep from accidentally introducing more calls to the Assert methods that don't work well.

AArnott:
Never mind. I guess the convenient place to wrap and rethrow that I hoped for isn't there. I'll take this PR with one stylistic change that I'll make myself. Thanks.

PR #14, opened 2017-06-07
Add NoFastCheck AssemblyNameComparer

### LINKED ISSUE #13: TypeRefs are considered equal if they are built by an assembly with the same path but a different assembly name
Repro:
```
AssemblyName a1 = CreateAssemblyName("AssemblyA", new Version(1, 0), CultureInfo.CurrentCulture, @"C:\some\path\AssemblyA.dll", publicKey: publicKey);
AssemblyName a2 = CreateAssemblyName("AssemblyA", new Version(2, 0), CultureInfo.CurrentCulture, @"C:\some\path\AssemblyA.dll");
Assert.NotEqual(TypeRef.Get(a1, ...), TypeRef.Get(a2, ...));
```

The reason is that `ByValueEquality+AssemblyNameComparer` returns true whenever the code bases are equal.

The scenario occurs when we compare the MEF cache and the newly created MEF graph to see if they are equivalent (if they are equivalent, we don't serialize the new graph to save time). Since the MEF cache was created with a different version of the assembly, VS MEF should not consider the code base "fast check" (the AssemblyName in the cache was created by a different process).

This is the root cause of feedback item 447376 in DevDiv.

AlexEyler:
Fixes #13.

This adds an "AssemblyNameNoFastCheck" EqualityComparer that will not compare the code bases when checking for equality (the code bases were used to quickly end the equality check since it was assumed that two AssemblyNames with the same code base are equal). In order to fix the scenario in #13, TypeRef.Equals is modified to use the AssemblyNameNoFastCheck comparer. This also changes the equality comparer in FieldRef to use the AssemblyNameNoFastCheck comparer.


src/Microsoft.VisualStudio.Composition/ByValueEquality+AssemblyNameComparer.cs
```diff
@@ -32,8 +40,8 @@ public bool Equals(AssemblyName x, AssemblyName y)
                     return true;
                 }
 
-                // fast path
-                if (x.CodeBase == y.CodeBase)
+                // If fast check is enabled, we can compare the code bases
+                if (this.fastCheck && x.CodeBase == y.CodeBase)
```
AArnott: Who still depends on the 'fast' check? Do they really need it to be fast, or can we simplify by removing this fast check altogether?

src/tests/Microsoft.VisualStudio.Composition.Tests/ByValueEqualityTests.cs
```diff
@@ -79,6 +79,17 @@ public void AssemblyNameComparer_ReturnsFalseWhenPublicKeysDontMatch()
             Assert.NotEqual(a1, a2, ByValueEquality.AssemblyName);
         }
 
+        [Fact]
+        public void AssemblyNameComparerNoFastCheck_SkipsCodeBaseCheck()
+        {
+            byte[] publicKey = GetPublicKeyFromExecutingAssembly();
+            AssemblyName a1 = CreateAssemblyName("AssemblyA", new Version(1, 0), CultureInfo.CurrentCulture, @"C:\some\path\AssemblyA.dll", publicKey: publicKey);
+            AssemblyName a2 = CreateAssemblyName("AssemblyA", new Version(2, 0), CultureInfo.CurrentCulture, @"C:\some\other\path\AssemblyA.dll");
```
AArnott: You also gave them different paths. So it seems you're not really testing the new functionality. Did you mean for these paths to be the same, since you're testing different assembly versions?

src/Microsoft.VisualStudio.Composition/ByValueEquality+AssemblyNameComparer.cs
AlexEyler: It's used as a comparer for some dictionaries (e.g input assemblies). I was hesitant to remove the fast check before knowing the potential perf impact (I'm trying to get a val build with these changes in but it's proving to be a bit difficult - I can't uses the vs-mef build definition since my branch does not exist in this repo).

src/tests/Microsoft.VisualStudio.Composition.Tests/ByValueEqualityTests.cs
AlexEyler: Whoops yes definitely, good call.

src/Microsoft.VisualStudio.Composition/ByValueEquality+AssemblyNameComparer.cs
AArnott: why not push your branch to this repo?

AlexEyler: yeah that's what I'm going to do.

PR #21, opened 2017-06-16
Add support for .NET Standard 1.5

### LINKED ISSUE #15: Support for .NET Standard
I believe we should be able to target .NET Standard 1.5. 

Work is proceeding in the [dev/andarno/netstandard](https://github.com/Microsoft/vs-mef/tree/dev/andarno/netstandard) branch.

* [X] Get the library compiling for .NET Standard
* [X] Get all test dependencies to build on .NET Standard
* [x] Get the test project itself to build on .NET Standard
* [x] Get the tests to pass on .NET Standard

AArnott:
This adds .netstandard1.5 and netcoreapp1.0 as targets for VS MEF.

Unit tests run on net45 as before, and now also runs on netcoreapp1.0 to prove that vs-mef works on both platforms.

A couple of caveats regarding .NET Core support:
* slow reflection only (by name rather than by metadata token)
* NoPIA isn't supported on CoreCLR

Closes #15


KirillOsenkov:
It all looks good to me as far as I can see.

src/Microsoft.VisualStudio.Composition/StandardAssemblyLoader.cs
```diff
@@ -31,6 +31,19 @@ public Assembly LoadAssembly(AssemblyName assemblyName)
 
             if (assembly == null)
             {
+#if NETCOREAPP1_0
+                // Clone the AssemblyName to undefine the CodeBase property in case it is set.
+                // Workaround for https://github.com/dotnet/coreclr/issues/10561
+                assemblyName = new AssemblyName
```
AlexEyler: nit: consider doing the same method as the fix for that issue here (see https://github.com/dotnet/coreclr/pull/11010/files#diff-941b565c09cd92cf3c9588857554f825R108).

AArnott: I considered that, but can't because Clone and CodeBase are internal I think.

PR #28, opened 2017-08-04
Speed up cost of reflection a bit

AArnott:
We cache the result of reflection at a few easy places.

More can certainly be done here and will be included in subsequent PRs.

I also take the opportunity to fix Equals and GetHashCode methods in light of the dual mode (metadata token vs. actual Reflection info object).


src/Microsoft.VisualStudio.Composition/Reflection/ResolverExtensions.cs
```diff
@@ -182,32 +178,9 @@ public static MemberInfo Resolve(this MemberRef memberRef)
             throw new NotSupportedException();
         }
 
+        [Obsolete("Use " + nameof(MemberRef) + " instead.", error: true)]
         public static MemberInfo Resolve(this MemberDesc memberDesc)
```
AArnott: This method was always just broken. And we never used it. So rather than fix it (which would require a breaking change), since it was never used I just scrape out its implementation. I'll remove these broken and unused types in v16.

PR #32, opened 2017-08-28
Add docs on substitutes for dynamic recomposition

AArnott:


doc/dynamic_recomposition.md
```diff
@@ -0,0 +1,60 @@
...
+
+At runtime in your test, set up your VS-MEF catalog to include `MockSupplier` along with the product assembly parts that import `IFoo` and `IBar`. Then with the runtime `ExportProvider`, initialize the mocks:
+
+```csharp
+[Fact]
+public void SomeTest()
+{
+    ExportProvider ep; // TODO: initialize this.
+
+    // Set up the mocks before acquiring other exports.
```
IlyaBiryukov: So, if other exports tried to get IFoo or IBar first before the test sets them up, what's going to happen?

doc/dynamic_recomposition.md
```diff
@@ -0,0 +1,60 @@
+# Dynamic Recomposition
+
+VS-MEF does not support .NET MEF's dynamic recomposition feature. This feature is exposed in .NET MEF in several ways, including:
```
IlyaBiryukov: What's going to happen with ImportAttribute.AllowRecomposition property? Is it going to be ignored?

doc/dynamic_recomposition.md
AArnott: When the test queries for an export that requires activating a part that imports an `IFoo`, then MEF would activate the `MockSupplier` and query its property. If you have already set the property's value, the MEF import will be satisfied with that value; otherwise the default value for that property will be used to satisfy the import (null in this case).

AArnott: I've added more comments to this doc to help explain this.

doc/dynamic_recomposition.md
AArnott: Yes, we ignore that parameter. it's fine if they support it -- we just don't ever do it.

PR #36, opened 2017-10-01
Fixes handling of generic type argument in importing constructor

AArnott:
We had some code that seemed to try to support this, but in fact open generic imports are not allowed and tests exists to prevent it, but none tested generic type parameters used in an importing constructor. This change adds just that, and then fixes up the library to reject it properly.

This change also prepares for allowing importing constructors to actually be static factory methods, which we may need later for a stable cache based on a compiled assembly.


AArnott:
I'm going to pull this back till I can separate out the importing constructor factory method bit, which is unrelated and isn't accompanied by tests.

PR #37, opened 2017-10-09
fix broken link in .md file

lukka:
fix broken link in .md file


AArnott:
Thanks, @lukka. Would you mind re-targeting your change to the v15.5 branch since the problem exists there as well? We'll merge v15.5 into master after that to get your fix here too.

lukka:
retargeting v15.5

PR #46, opened 2017-11-10
Add support for metadata view classes with default constructors

nicolaihenriksen:
MEFv2 (i.e. "NuGet MEF") supports metadata views defined by classes with a public default constructor and public property setters. VS MEF was missing this feature (probably because MEFv1 doesn't share it). 

This change adds this feature with a bunch of tests for variations on the idea. Some tests verify correct configurations and other tests verify incorrect configurations fail the way we would expect.

One slight deviation from the MEFv2 feature is that VS MEF supports internal types used as metadata views. This allows libraries to use VS MEF internally without exposing implementation details.

### DIFF
--- a/src/Microsoft.VisualStudio.Composition/MetadataViewClassDefaultCtorProvider.cs
+++ b/src/Microsoft.VisualStudio.Composition/MetadataViewClassDefaultCtorProvider.cs
@@ -37,6 +37,7 @@ namespace Microsoft.VisualStudio.Composition
         public object CreateProxy(IReadOnlyDictionary<string, object> metadata, IReadOnlyDictionary<string, object> defaultValues, Type metadataViewType)
         {
             Requires.NotNull(metadata, nameof(metadata));
+            Requires.NotNull(defaultValues, nameof(defaultValues));
             Requires.NotNull(metadataViewType, nameof(metadataViewType));
 
             TypeInfo typeInfo = metadataViewType.GetTypeInfo();
--- a/src/Microsoft.VisualStudio.Composition/MetadataViewClassDefaultCtorProvider.cs
+++ b/src/Microsoft.VisualStudio.Composition/MetadataViewClassDefaultCtorProvider.cs
@@ -44,7 +44,7 @@ namespace Microsoft.VisualStudio.Composition
 
             foreach (var propertyInfo in typeInfo.GetProperties(BindingFlags.Instance | BindingFlags.Public))
             {
-                if (metadata.TryGetValue(propertyInfo.Name, out object value) && propertyInfo.SetMethod != null)
+                if ((metadata.TryGetValue(propertyInfo.Name, out object value) || defaultValues.TryGetValue(propertyInfo.Name, out value)) && propertyInfo.SetMethod != null)
                 {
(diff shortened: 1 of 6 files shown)


nicolaihenriksen:
I should add that a test for [ImortMany] and a collection of lazy parts with strongly typed metadata should be added as well. This just illustrates the issue.

src/Microsoft.VisualStudio.Composition/MetadataViewClassDefaultCtorProvider.cs
```diff
@@ -0,0 +1,63 @@
...
+        {
+            Requires.NotNull(metadataType, nameof(metadataType));
+            var typeInfo = metadataType.GetTypeInfo();
+
+            return typeInfo.IsClass && !typeInfo.IsAbstract && FindConstructor(typeInfo) != null;
+        }
+
+        public object CreateProxy(IReadOnlyDictionary<string, object> metadata, IReadOnlyDictionary<string, object> defaultValues, Type metadataViewType)
+        {
+            Requires.NotNull(metadata, nameof(metadata));
```
jepetty: If we check that `metadata` is not null, should we do the same for `defaultValues`? Or, because as defined in the interface they can never be null, can we avoid the check altogether?

src/tests/Microsoft.VisualStudio.Composition.Tests/MetadataViewClassDefaultCtorProviderTests.cs
```diff
@@ -0,0 +1,181 @@
...
+        public void PublicCtorInternalProperty(IContainer container)
+        {
+            var part = container.GetExportedValue<DefaultMetadataViewImporter>();
+            Assert.Null(part.InnerPart.Metadata.InternalProperty);
+        }
+
+        [MefFact(CompositionEngines.V2Compat, typeof(MetadataDecoratedPart), typeof(InternalCtorMetadataViewImporter), InvalidConfiguration = true)]
+        public void InternalCtorPublicProperty(IContainer container)
+        {
+            var part = container.GetExportedValue<InternalCtorMetadataViewImporter>();
```
jepetty: Is this test testing something?

AArnott: Yes, although I can see how it's not clear. In the `MefFact` attribute on this test method, there is this property set: `InvalidConfiguration = true`. This asserts that either during composition or during test method execution that a MEF exception is thrown.

src/Microsoft.VisualStudio.Composition/MetadataViewClassDefaultCtorProvider.cs
AArnott: I missed that. I'll add the `defaultValues` check. 
The interface xml doc is just that: documentation. It makes it 'legal' for an implementation to throw if the contract is violated. I generally code defensively like this because NullRefExceptions are just too painful to debug later. I only skip input/result verification when it's a concrete type that I have a _lot_ of trust in (like BCL types).

PR #48, opened 2017-12-07
Support inheritance on MEFv2's [MetadataAttribute]

nicolaihenriksen:
In our setup we use this type of construct via Nuget MEF to have shared stuff on the base export attribute.

You can argue that we could do this differently, but nonetheless it works in Nuget MEF :-)

If you have questions or comments, please contact me.

### DIFF
--- a/src/Microsoft.VisualStudio.Composition/Configuration/AttributedPartDiscovery.cs
+++ b/src/Microsoft.VisualStudio.Composition/Configuration/AttributedPartDiscovery.cs
@@ -243,8 +243,8 @@ namespace Microsoft.VisualStudio.Composition
                     // Perf optimization, relies on short circuit evaluation, often a property attribute is an ExportAttribute
                     if (attrType != typeof(ExportAttribute).GetTypeInfo() && attrType.IsAttributeDefined<MetadataAttributeAttribute>(inherit: true))
                     {
-                        var properties = attrType.GetProperties(BindingFlags.Public | BindingFlags.NonPublic | BindingFlags.Instance | BindingFlags.DeclaredOnly);
-                        foreach (var property in properties)
+                        var properties = attrType.GetProperties(BindingFlags.Public | BindingFlags.NonPublic | BindingFlags.Instance);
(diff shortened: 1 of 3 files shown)


nicolaihenriksen:
BTW I can see the commit comment on my second commit was a bit misleading. I meant to say "Exports with custom export attribute..."

AArnott:
@nicolaihenriksen feel free to force push to correct the commit if you want. I won't be building on your commits at *least* for several hours, and likely a couple days.

AArnott:
So interestingly enough, it turns out that the crux of the reason for your test to fail is that your derived custom metadata attribute doesn't have a `[MetadataAttribute]` attribute affixed to it. MEFv2 didn't require it, but VS MEF did. Adding that attribute to the derived custom metadata class works around the problem.

I've fixed the shortcoming though, so VS MEF with this PR will support what you were trying to do. But I removed your tests in favor of a single line change to our existing tests to exercise this feature.

While at it, I noticed your comment about how MEFv2 doesn't support properties defined on base custom metadata classes. I also noticed we already had a test for this, but it was disabled for VS MEF because VS MEF did not share that shortcoming. However, since the goal is for VS MEF to read MEFv2 attributes the same way MEFv2 would (with a few exceptions) I also "fixed" VS MEF to have the same shortcoming, and enabled the test we already had to assert this.

AArnott:
Meh, I don't care too much for emulating MEFv2's limitation regarding properties on custom metadata attribute base types. I'm going to keep VS MEF superior in that regard.

nicolaihenriksen:
@AArnott  Great!! Is there a CI build available now, or do I have to wait until the PR is completed?

I guess I can just add the `[MetadataAttribute]` attribute and go from there until I have a new CI package to work with...

AArnott:
> Is there a CI build available now, or do I have to wait until the PR is completed?

Not yet. We have AppVeyor set to not publish packages for PRs. I expect this PR will complete in the next couple business days.

But ya, you can just add the `[MetadataAttribute]` to workaround it for now.

PR #49, opened 2017-12-10
Multi-target to net46

AArnott:
This gets us ready to be able to code up optimizations that only work on net46


AArnott:
Apparently the failure to upload code coverage data is unrelated to this change -- it's failing in master already.

PR #54, opened 2017-12-19
Remove LastWriteTimeUtc from StrongAssemblyIdentity

AArnott:
Per @KirillOsenkov's comments in #53, the MVID should be a sufficiently unique value to confirm the assembly hasn't changed.

### DIFF
--- a/src/Microsoft.VisualStudio.Composition/Resolver.cs
+++ b/src/Microsoft.VisualStudio.Composition/Resolver.cs
@@ -80,18 +80,15 @@ namespace Microsoft.VisualStudio.Composition
                 assemblyName = assembly.GetName();
             }
 
-            lock (this.loadedAssemblyStrongIdentities)
+            if (this.TryGetAssemblyId(assemblyName, out StrongAssemblyIdentity result))
             {
-                if (this.loadedAssemblyStrongIdentities.TryGetValue(assemblyName, out var result))
-                {
-                    return result;
-                }
+                return result;
             }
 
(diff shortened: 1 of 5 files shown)


src/Microsoft.VisualStudio.Composition/Resolver.cs
```diff
@@ -80,18 +80,15 @@ private StrongAssemblyIdentity NotifyAssemblyLoaded(Assembly assembly, AssemblyN
...
-                    return result;
-                }
+                return result;
             }
 
             var assemblyId = StrongAssemblyIdentity.CreateFrom(assembly, assemblyName);
             lock (this.loadedAssemblyStrongIdentities)
             {
-                if (!this.loadedAssemblyStrongIdentities.TryGetValue(assemblyName, out var result))
+                if (!this.loadedAssemblyStrongIdentities.TryGetValue(assemblyName, out result))
```
AlexEyler: I'm not too familiar with the code here - why not use TryGetAssemblyId here as well?

AArnott: Because this isn't just a check within a lock: it's a check-then-initialize multi-step that all has to be within a lock. So calling the other method would not take a lock with a broad enough scope, resulting in two locks, which while still cheap, makes the helper method also worth next to nothing.

AlexEyler: Makes sense to me, thanks!

PR #56, opened 2017-12-22
Avoid resolving type in RuntimeImport.IsLazy

AArnott:
While inspecting a MEF graph, checking whether an import is lazy currently require loading the assemblies that define the imported type. 

This change *probably* doesn't change anything meaningful (which is why I didn't add any tests to verify it), since the assembly will have to be loaded a moment later anyway as part of setting the importing property. But this does make our API a bit more consistent in that almost everything deals with TypeRef's instead of Type's. And the new `IsAnyLazyType` extension method that is added in this PR creates a "pit of success" for lazy type checks instead of currently leading folks into writing code that will have to load assemblies in other places.


src/Microsoft.VisualStudio.Composition/LazyServices.cs
```diff
@@ -40,6 +43,24 @@ internal static bool IsAnyLazyType(this Type type)
             return false;
         }
 
+        /// <summary>
+        /// Gets a value indicating whether a type is a Lazy`1 or Lazy`2 type.
```
jepetty: What is the difference between a Lazy1 and a Lazy2 type? And _nit_ can we change to "Lazy1 or Lazy2 type, or neither"? I was confused that the return would indicate that it was a Lazy1 type or it was a Lazy2 type.

AArnott: I've updated the xml docs to make this clear. Thanks.
FYI: The Lazy`1 is the CLR type syntax for generics. A generic type has a backtick followed by the number of generic type arguments it accepts. So there are two Lazy types: one takes T and one takes both T and TMetadata.

PR #57, opened 2017-12-28
OnImportsSatisfied call order

nicolaihenriksen:
In MEFv1, it seems that if type A contains an instance of type B and both types implement IPartImportsSatisfiedNotification, the OnImportsSatisfied() method is called on type B before type A.

I am not convinced that this is always the case (circular dependencies?!) or just coincidence/luck. Anyways, here is a test that passes in v1 and fails in v2 and v3.

Personally I think having a dependency on this call order in our code is horrific, so I will look into refactoring that. But I thought you should have a heads up just so you can decide whether this is something you need to do something about in vs-mef.

### DIFF
--- a/src/tests/Microsoft.VisualStudio.Composition.Tests/OnImportsSatisfiedTests.cs
+++ b/src/tests/Microsoft.VisualStudio.Composition.Tests/OnImportsSatisfiedTests.cs
@@ -29,8 +29,17 @@ namespace Microsoft.VisualStudio.Composition.Tests
             Assert.Equal(1, part.ImportsSatisfiedInvocationCount);
         }
 
-        [MefFact(CompositionEngines.V1Compat | CompositionEngines.V2Compat, typeof(OuterPart), typeof(InnerPart))]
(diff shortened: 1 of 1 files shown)


nicolaihenriksen:
I have been trying to read up on the call order in MEFv1, but I cannot find anything explicitly mentioning it. However the link below does actually give a hint about it in the sentence:

> `IPartImportsSatisfiedNotification` contains one method named `OnImportsSatisfied`. This method is called by the composition container on any parts that implement the interface when composition has been completed and **the part's imports are ready for use**.

https://docs.microsoft.com/en-us/dotnet/framework/mef/attributed-programming-model-overview-mef

AArnott:
Thanks for your submission, @nicolaihenriksen. From the .NET MEF source code and docs, I think this is an undocumented behavior. What's more, it would reverse itself if the outer part imported the inner part using `Lazy<T>`, which is often a preferred path to avoid initializing too much of the graph unnecessarily. So I've adjusted your test so as to serve as documentation for .NET MEF behavior but we will not strive to emulate that behavior in VS MEF. More explanation is in the code comment I added to your test.

Thanks again.

PR #58, opened 2017-12-31
Reduce memory consumption

### LINKED ISSUE #30: Heap regression in master branch
* [x] Check that we are interning `string` and `AssemblyName` instances when deserializing from the cache.
* [ ] Convert the `*Ref` structs to classes as they are significantly larger than 16 bytes and they appear in many other types (transitively or directly).

AArnott:
Take the hit of some API breaking changes in order to reduce memory consumption by changing some structs to classes. These were large structs and some were 'stored' in several places. By using classes instead of structs, these duplicates will be represented in memory only once.

The breaking changes are to the API that is mostly just used internally anyway, as attested to by the very few unit test changes that had to be made (we haven't bothered writing tests for these APIs). If there was an externally written `PartDiscovery`-derived class, that would be significantly impacted by this change. But I've never heard of one existing and it seems very unlikely.

Since we're taking a hit of breaking changes, I went ahead and removed some `[Obsolete]` APIs as well, some of which were completely unused anywhere. We also see a bunch of code removed that was trying to fake up polymorphism for the `MemberRef` and related structs since now we can actually use the real thing since they are classes.

Fixes #30 

* [x] Run RPS tests on this change prior to checkin to measure impact.

image
image
image
image


AArnott:
* [ ] Run RPS tests on this change prior to checkin to measure impact.

AArnott:
@ikeras may be interested to see [this earlier regression](https://devdiv.visualstudio.com/DevDiv/_workitems/edit/476150) corrected in 15.8  even though we had obtained a permanent exception.

src/tests/Microsoft.VisualStudio.Composition.Tests/ComposableCatalogTests2.cs
```diff
@@ -302,6 +302,17 @@ public void GetAssemblyInputs_Empty()
...
 
+        [Theory]
+        [MemberData(nameof(DiscoveryEnginesTheoryData))]
+        public async Task GetAssemblyInputs_WithImports(PartDiscovery discovery)
+        {
+            var catalog = TestUtilities.EmptyCatalog.AddParts(
+                await discovery.CreatePartsAsync(
+                    typeof(ExportingType),
+                    typeof(ImportingType)));
+            catalog.GetInputAssemblies();
```
AlexEyler: This test isn't asserting anything - should you call this.AssertExpectedInputAssemblies? Are is this just testing to make sure there aren't any exceptions thrown when calling GetInputAssemblies?

src/tests/Microsoft.VisualStudio.Composition.Tests/XunitExtensions/CatalogScrambler.cs
```diff
@@ -212,33 +190,18 @@ private PropertyRef Scramble(PropertyRef propertyRef)
...
-                return new MemberRef(this.Scramble(importingMemberRef.Type));
-            }
-            else if (importingMemberRef.IsEmpty)
-            {
-                return default(MemberRef);
-            }
-            else
-            {
-                throw new NotImplementedException();
+            switch (importingMemberRef)
```
AlexEyler: This is so much prettier :)

src/tests/Microsoft.VisualStudio.Composition.Tests/ComposableCatalogTests2.cs
AArnott: Right: `GetInputAssemblies()` was throwing an NRE because we had no test for importing types in this class. So I added one, and this test just includes that and calls the method to ensure it doesn't throw.

PR #63, opened 2018-01-20
Enable .NET MEF attribute support on .NETStandard

AArnott:
This exposes the .NETStandard compliant .NET MEF attributes we were already using as a compilation aid for our unit tests as publishable packages, and takes a dependency on these packages in VS MEF itself so that folks can use the .NET MEF attributes in .NET Standard projects and have it work on both .NET Core and .NET Framework. 

These .NET MEF attribute assemblies are modeled after the MEFv2 attribute factoring (distinguishing attributes versus runtime types across two assemblies). When these assemblies compile for .NET Framework they are simply type forwarders to the .NET types, but they carry the (minimal) code that they require to stand on their own when compiled for .NET Standard. This ensures that assemblies can compile against .NET Standard or .NET Framework versions of these assemblies and run properly as expected on compatible runtimes.
In addition, given that .NET MEF evolved in .NET 4.5 over its initial .NET 4.0 version, these new assemblies omit the 4.5-only types when compiled for net40, as appropriate.

Once VS MEF compiled against these attribute assemblies, it became possible to support part discovery on these attributes for all target frameworks. I then changed the unit test infrastructure to add coverage of v1 attributes when running on .NET Core, and found just one bug that could not be reached until this was enabled.

**Notes for code reviewers:**

The movement of the System.ComponentModel.Composition.Runtime and System.ComponentModel.Composition.AttributedModel projects out of the test subdirectory includes small changes including new xml doc comments and adding few missing members. The github diff shows I added/removed entire files in a couple cases which is not accurate.

Code coverage on .NET Core goes up immensely with this change. Before this change only 1194 tests ran on .NET Core (out of 2320 on .NET Framework). After this change, 1831 (+637) tests run on .NET Core.


AArnott:
I spoke with the .NET team folks who own this area back on 6/14/2017 regarding this plan, and they had no objections.

CC: @terrajobst @danmosemsft @karelz

PR #65, opened 2018-02-17
Run tests on .NET Core 2.0

AArnott:
No changes to the product in this case, but a couple tests had to be fixed up for WeakReference behavior that is unique to .NET Core 2.0 as documented at https://github.com/dotnet/coreclr/issues/13490

### DIFF
--- a/.appveyor.yml
+++ b/.appveyor.yml
@@ -62,9 +62,9 @@ test_script:
 
     cd src\tests\Microsoft.VisualStudio.Composition.Tests
 
-    dotnet test -f netcoreapp1.0 --no-build
+    dotnet test -f netcoreapp1.0 --no-build --no-restore
 
-    dotnet test -f netcoreapp2.0 --no-build
+    dotnet test -f netcoreapp2.0 --no-build --no-restore
 artifacts:
 - path: bin\%configuration%\Packages\Microsoft.VisualStudio.Composition.15.*.nupkg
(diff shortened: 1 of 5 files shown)


AArnott:
@KirillOsenkov Given the two tests I modified are both excluded from testing on mono, I wonder if it's for the same reason and one that I just addressed for .NET Core 2.0. Are you interested in testing on mono with these methods enabled?

AArnott:
@KirillOsenkov When I run the net46 tests on mono using simply "mono xunit.console.exe net46\tests.dll" I get several test failures due to what appears to be mono's implementation of MEFv1. Are you seeing those too?

KirillOsenkov:
Is this mono on Windows or on Mac? What is `mono --version`? The mono runtime team has fixed a whole bunch of bugs over the past year that were blocking, I haven't checked if all of them are now fixed in the latest version.

AArnott:
This is the latest that their web site had to install yesterday.

```
mono --version
Mono JIT compiler version 5.8.0 (Visual Studio built mono)
Copyright (C) 2002-2014 Novell, Inc, Xamarin Inc and Contributors. www.mono-project.com
        TLS:           normal
        SIGSEGV:       normal
        Notification:  Thread + polling
        Architecture:  x86
        Disabled:      none
        Misc:          softdebug
        LLVM:          supported, not enabled.
        GC:            sgen
```

PR #75, opened 2018-03-29
Add travis-ci for osx and linux

dgriffen:
Adds support for CI runs for osx and linux via travis-ci. This is just a simple initial configuration and only tests against .NET Core 2.0. Actual builds will have to be enabled in the repo by an admin.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,0 +1,17 @@
+language: csharp
+mono: none
+dotnet: 2.1.4
+
+os:
+  - linux
+  - osx
+
+git:
+  depth: false
+
+install:
+- dotnet restore src/Microsoft.VisualStudio.Composition.sln
+
+script:
+- dotnet build -f netcoreapp2.0 src/Microsoft.VisualStudio.Composition/
+- dotnet test -f netcoreapp2.0 --no-restore src/tests/Microsoft.VisualStudio.Composition.Tests/
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,6 +1,8 @@
 language: csharp
 mono: none
 dotnet: 2.1.4
+mono: 5.10.0
+
 branches:
   only:
   - master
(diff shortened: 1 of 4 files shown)


AArnott:
What does an admin need to do to convince travis-ci to build this?

.travis.yml
```diff
@@ -0,0 +1,17 @@
+language: csharp
+mono: none
+dotnet: 2.1.4
```
AArnott: Can we filter this down to the same subset of branches that our appveyor.yml file expresses?

AArnott:
I kicked one off: https://travis-ci.org/Microsoft/vs-mef/builds/360103568

.travis.yml
AArnott: I pushed a fix for this myself.

AArnott:
It would certainly be interesting to add a configuration for travis-ci that runs tests on mono since that powers Visual Studio for Mac.

AArnott:
I guess I got the Travis-CI integration going already, since I'm seeing it validating this PR. :)

AArnott:
Are there any other settings, ignored files, etc. that are in appveyor.yml for which we can set equivalents in travis-ci.yml?

dgriffen:
@AArnott probably, I'll take a look at the spec and see what ignores I can put into it. I will also add tests for mono on the osx build.

AArnott:
awesome. I'm OK if you complete this PR in the meantime and follow up with any more CI switches or mono tests in a subsequent PR.

AArnott:
And BTW, I don't expect all the tests to pass on mono. @KirillOsenkov may know the latest on this. But we'll want to rig a test run on mono that matches what we *expect* to pass so that we can recognize regressions. We already have a way to do this filtering of tests on mono -- but the set of tests we run on it may be out of date.

AArnott:
> What version of Mono is used in CI?

@KirillOsenkov I suppose that's up to us. Which version do you want us to test with? Presumably the version that VS 4 Mac uses.

dgriffen:
Turns out travis-ci does not have support for ignoring files like appveyor does (see [#6301](https://github.com/travis-ci/travis-ci/issues/6301)).

KirillOsenkov:
Minimum required version for VSMac is currently 5.12.0.98.

REQUIRED_MONO_URL_PUBLIC=https://xamjenkinsartifact.azureedge.net/build-package-osx-mono/2018-02/87/ed56e90aca9298121c8edabe200193ae861332af/MonoFramework-MDK-5.12.0.98.macos10.xamarin.universal.pkg

dgriffen:
@KirillOsenkov Where is mono 5.12 supposed to come from? According to [mono-project.com](http://www.mono-project.com/) the latest version is 5.10

dgriffen:
Looks like the `CacheStaleFromRecompiledAssembly` test fails on mono. Not sure exactly why but I assume its for the same reason other tests are skipped on mono due to AppDomain load differences.

KirillOsenkov:
Ah, right, yeah just take the latest publicly available.

dgriffen:
@AArnott should I skip the one failing test for now or should I leave it failing?

AArnott:
@dgriffen Yes, let's just mark the test to skip on Mono. I think the way we're doing that is by calling `           SkipOnMono();` within the test and using a `SkippableFact` attribute on the method.

AArnott:
@dgriffen I filed https://github.com/Microsoft/vs-mef/issues/77 to track one of the test failures in your latest Travis-CI build that appears unrelated to anything you've been doing here, but you happened to hit a race condition in one of the tests.

PR #83, opened 2018-04-18
Improve tracing of serialization

AArnott:
- Use `nameof` for serialization trace messages
- Fix StyleCop issue when TRACESERIALIZATION is defined
- Fix serialization trace messages for lists
- Add object reuse info to serialization trace
- Also trace to files instead of Debug window for easier diff'ing during debugging.

### DIFF
--- a/src/Microsoft.VisualStudio.Composition/Configuration/SerializationContextBase.cs
+++ b/src/Microsoft.VisualStudio.Composition/Configuration/SerializationContextBase.cs
@@ -157,7 +157,7 @@ namespace Microsoft.VisualStudio.Composition
 
         protected void Write(MethodRef methodRef)
         {
-            using (this.Trace("MethodRef"))
+            using (this.Trace(nameof(MethodRef)))
             {
                 if (methodRef == null)
                 {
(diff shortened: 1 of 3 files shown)


src/Microsoft.VisualStudio.Composition/Configuration/SerializationContextBase.cs
```diff
@@ -141,6 +155,16 @@ protected internal void FinalizeObjectTableCapacity()
...
+        {
+            if (disposing)
+            {
+#if TRACESERIALIZATION || TRACESTATS
+                this.trace.Dispose();
+#endif
+            }
+        }
+
         protected SerializationTrace Trace(string elementName, bool isArray = false)
```
dgriffen: Should this be changed to use `TypeRefFlags` once that PR is in?

AArnott: What are you referring to? The code location of this comment doesn't seem relevant to `TypeRefFlags`.

dgriffen: I guess I figured that the isArray stuff was related to `TypeRefFlags` from #85 since that used to only have isArray. My thinking was that since the `Trace` method accepts isArray is should be expanded to instead take the more general `TypeRefFlags` parameter.

AArnott: No, this is a completely different `isArray` flag. This wants to know if literally we're serializing an array, which might be of bool or any other type. `TypeRefFlags` is about serializing a single `TypeRef` element.

PR #85, opened 2018-04-18
Allow TypeRef to store a variety of flags

AArnott:
This should make our public API more resilient to future changes (so it's not a breaking change every time we add a boolean flag.


AArnott:
@dgriffen Great question. This is not the first nor the largest API breaking change in 15.8. So for 15.8 I think it's already a past issue. But going forward, prioritizing semver conformance over matching the VS version is possible, but IMO given we match VS today, it would make it very confusing for VS customers to know which library version ships with which VS version. We do have some non-VS users, true. But VS is still the largest customer. 
Also considering these API breaking changes aren't expected to be user impacting (since they're all in reflection/discovery APIs), IMO we should continue matching VS versions at least until customers actually complain that we're breaking them without warning.

AArnott:
@dgriffen If we're not too late for 15.8, can you complete this PR and arrange a VS MEF insertion so we can include this and fix #88?

dgriffen:
I'll merge this and then see what the status is for 15.8 feature work on mef.

PR #89, opened 2018-05-04
Fix support for Dictionary<string, object> metadata views

### LINKED ISSUE #88: Importing metadata with Dictionary<string,object> broken
Importing with the concrete type instead of the interface seems to have broken in the master branch (it works in v15.6):

```cs
[Import, MefV1.Import]
public Lazy<PartWithExportMetadata, Dictionary<string, object>> ImportingProperty { get; set; }
```

AArnott:
In #46 we added support for metadata view classes with default constructors. We set the order precedence such that this metadata view provider is more preferred than the one that instantiates metadata view classes by passing in a dictionary as an argument to the constructor.

But it is this now-less preferred provider that *was* supporting `Dictionary<string,object>` metadata views (for which we had *no* tests 😲). The new, and more preferred, metadata view provider was noticing that `Dictionary<K,V>` has a default ctor so it was happy to claim it as supported, but of course `Dictionary<K,V>` has no settable properties, and certainly not the ones that the client is expecting. So it effectively hid all metadata.

This fix fixes the preferred order so that the original metadata view provider (the one that takes dictionaries as a ctor parameter) is more preferred over the metadata view provider that works with default ctors. This not only fixes the problem, but in the event both actually were correct in working, this ordering is more efficient because it means we won't rely on reflection unnecessarily.

Fixes #88


AArnott:
GitHub is malfunctioning so I can't add @dgriffen or @wade0016 as reviewers.

PR #93, opened 2018-05-08
Fix support for multiple internal metadata view interfaces defined from multiple assemblies

### LINKED ISSUE #91: Internal interface metadata view proxy generation fails after first assembly
As added in #10, VS MEF claims to support internal metadata view interfaces by telling the CLR to skip visibility checks. We have a test for this, but it turns out the CLR fails when asked to skip visibility checks for the second assembly.

```
at System.Reflection.Emit.TypeBuilder.TermCreateClass(RuntimeModule module, Int32 tk, ObjectHandleOnStack type)
   at System.Reflection.Emit.TypeBuilder.CreateTypeNoLock()
   at System.Reflection.Emit.TypeBuilder.CreateTypeInfo()
   at Microsoft.VisualStudio.Composition.MetadataViewGenerator.GenerateInterfaceViewProxyType(Type viewType) in D:\git\MEF\src\Microsoft.VisualStudio.Composition\Configuration\MetadataViewGenerator.cs:line 263
   at Microsoft.VisualStudio.Composition.MetadataViewGenerator.GetMetadataViewFactory(Type viewType) in D:\git\MEF\src\Microsoft.VisualStudio.Composition\Configuration\MetadataViewGenerator.cs:line 127
   at Microsoft.VisualStudio.Composition.ExportMetadataViewInterfaceEmitProxy.MetadataViewProxyProvider.CreateProxy(IReadOnlyDictionary`2 metadata, IReadOnlyDictionary`2 defaultValues, Type metadataViewType) in D:\git\MEF\src\Microsoft.VisualStudio.Composition\Configuration\ExportMetadataViewInterfaceEmitProxy.cs:line 34
   at Microsoft.VisualStudio.Composition.ExportProvider.GetStrongTypedMetadata(IReadOnlyDictionary`2 metadata, Type metadataType) in D:\git\MEF\src\Microsoft.VisualStudio.Composition\ExportProvider.cs:line 550
   at Microsoft.VisualStudio.Composition.RuntimeExportProviderFactory.RuntimeExportProvider.GetValueForImportElement(RuntimePartLifecycleTracker importingPartTracker, RuntimeImport import, RuntimeExport export, Func`3 lazyFactory) in D:\git\MEF\src\Microsoft.VisualStudio.Composition\RuntimeExportProviderFactory+RuntimeExportProvider.cs:line 237
   at Microsoft.VisualStudio.Composition.RuntimeExportProviderFactory.RuntimeExportProvider.GetValueForImportSite(RuntimePartLifecycleTracker importingPartTracker, RuntimeImport import) in D:\git\MEF\src\Microsoft.VisualStudio.Composition\RuntimeExportProviderFactory+RuntimeExportProvider.cs:line 133
   at Microsoft.VisualStudio.Composition.RuntimeExportProviderFactory.RuntimeExportProvider.RuntimePartLifecycleTracker.SatisfyImports() in D:\git\MEF\src\Microsoft.VisualStudio.Composition\RuntimeExportProviderFactory+RuntimeExportProvider.cs:line 549
   at Microsoft.VisualStudio.Composition.ExportProvider.PartLifecycleTracker.SatisfyImmediateImports() in D:\git\MEF\src\Microsoft.VisualStudio.Composition\ExportProvider.cs:line 1202
   at Microsoft.VisualStudio.Composition.ExportProvider.PartLifecycleTracker.MoveNext(PartLifecycleState nextState) in D:\git\MEF\src\Microsoft.VisualStudio.Composition\ExportProvider.cs:line 1263
   at Microsoft.VisualStudio.Composition.ExportProvider.PartLifecycleTracker.MoveToState(PartLifecycleState requiredState) in D:\git\MEF\src\Microsoft.VisualStudio.Composition\ExportProvider.cs:line 1335
   at Microsoft.VisualStudio.Composition.ExportProvider.PartLifecycleTracker.GetValueReadyToExpose() in D:\git\MEF\src\Microsoft.VisualStudio.Composition\ExportProvider.cs:line 1058
   at Microsoft.VisualStudio.Composition.ExportProvider.<>c__DisplayClass53_0.<CreateExport>b__0() in D:\git\MEF\src\Microsoft.VisualStudio.Composition\ExportProvider.cs:line 462
   at System.Lazy`1.CreateValue()
   at System.Lazy`1.LazyInitValue()
   at System.Lazy`1.get_Value()
   at Microsoft.VisualStudio.Composition.Export.get_Value() in D:\git\MEF\src\Microsoft.VisualStudio.Composition\Export.cs:line 54
   at Microsoft.VisualStudio.Composition.ExportProvider.<>c__DisplayClass70_1`2.<GetExports>b__1() in D:\git\MEF\src\Microsoft.VisualStudio.Composition\ExportProvider.cs:line 814
   at System.Lazy`1.CreateValue()
   at System.Lazy`1.LazyInitValue()
   at System.Lazy`1.get_Value()
   at Microsoft.VisualStudio.Composition.ExportProvider.GetExportedValue[T]() in D:\git\MEF\src\Microsoft.VisualStudio.Composition\ExportProvider.cs:line 256
   at Microsoft.VisualStudio.Composition.Tests.TestUtilities.V3ContainerWrapper.GetExportedValue[T]() in D:\git\MEF\src\tests\Microsoft.VisualStudio.Composition.Tests\TestUtilities.cs:line 715
   at Microsoft.VisualStudio.Composition.Tests.ExportMetadataTests.ThreadSafetyTestForInternalInterfaceProxyGeneration_2Then1(IContainer container) in D:\git\MEF\src\tests\Microsoft.VisualStudio.Composition.Tests\ExportMetadataTests.cs:line 161
```

AArnott:
This fixes our metadata view code gen so that when implementing internal interfaces, we generate a _new_ dynamic assembly each time a request comes in that requires a unique set of skip visiblity check attributes. When a request comes in that coincides with the set of skip check attributes of a dynamic assembly we've already opened, we reuse that one to keep the dynamic assembly count to a minimum.

Fixes #91

### DIFF
--- a/src/Microsoft.VisualStudio.Composition/Configuration/MetadataViewGenerator.cs
+++ b/src/Microsoft.VisualStudio.Composition/Configuration/MetadataViewGenerator.cs
@@ -79,7 +79,6 @@ namespace Microsoft.VisualStudio.Composition
         private const string MetadataViewFactoryName = "Create";
 
         private static readonly Dictionary<Type, MetadataViewFactory> MetadataViewFactories = new Dictionary<Type, MetadataViewFactory>();
-        private static readonly AssemblyName ProxyAssemblyName = new AssemblyName(string.Format(CultureInfo.InvariantCulture, "MetadataViewProxies_{0}", Guid.NewGuid()));
 
         private static readonly Type[] CtorArgumentTypes = new Type[] { typeof(IReadOnlyDictionary<string, object>), typeof(IReadOnlyDictionary<string, object>) };
         private static readonly MethodInfo MdvDictionaryTryGet = CtorArgumentTypes[0].GetTypeInfo().GetMethod("TryGetValue");
@@ -87,28 +86,12 @@ namespace Microsoft.VisualStudio.Composition
         private static readonly MethodInfo ObjectGetType = typeof(object).GetTypeInfo().GetMethod("GetType", Type.EmptyTypes);
         private static readonly ConstructorInfo ObjectCtor = typeof(object).GetTypeInfo().GetConstructor(Type.EmptyTypes);
 
-        private static ModuleBuilder transparentProxyModuleBuilder;
-        private static SkipClrVisibilityChecks skipClrVisibilityChecks;
-
         public delegate object MetadataViewFactory(IReadOnlyDictionary<string, object> metadata, IReadOnlyDictionary<string, object> defaultMetadata);
 
         private static AssemblyBuilder CreateProxyAssemblyBuilder(ConstructorInfo constructorInfo)
         {
-            return AssemblyBuilder.DefineDynamicAssembly(ProxyAssemblyName, AssemblyBuilderAccess.Run);
-        }
-
-        private static ModuleBuilder GetProxyModuleBuilder()
-        {
-            Assumes.True(Monitor.IsEntered(MetadataViewFactories));
-            if (transparentProxyModuleBuilder == null)
-            {
-                // make a new assemblybuilder and modulebuilder
-                var assemblyBuilder = CreateProxyAssemblyBuilder(typeof(SecurityTransparentAttribute).GetTypeInfo().GetConstructor(Type.EmptyTypes));
-                transparentProxyModuleBuilder = assemblyBuilder.DefineDynamicModule("MetadataViewProxiesModule");
-                skipClrVisibilityChecks = new SkipClrVisibilityChecks(assemblyBuilder, transparentProxyModuleBuilder);
-            }
-
-            return transparentProxyModuleBuilder;
+            var proxyAssemblyName = new AssemblyName(string.Format(CultureInfo.InvariantCulture, "MetadataViewProxies_{0}", Guid.NewGuid()));
+            return AssemblyBuilder.DefineDynamicAssembly(proxyAssemblyName, AssemblyBuilderAccess.Run);
         }
 
         public static MetadataViewFactory GetMetadataViewFactory(Type viewType)
@@ -141,7 +124,9 @@ namespace Microsoft.VisualStudio.Composition
             TypeBuilder proxyTypeBuilder;
             Type[] interfaces = { viewType };
 
-            var proxyModuleBuilder = GetProxyModuleBuilder();
+            var assemblyBuilder = CreateProxyAssemblyBuilder(typeof(SecurityTransparentAttribute).GetTypeInfo().GetConstructor(Type.EmptyTypes));
+            var proxyModuleBuilder = assemblyBuilder.DefineDynamicModule("MetadataViewProxiesModule");
+            var skipClrVisibilityChecks = new SkipClrVisibilityChecks(assemblyBuilder, proxyModuleBuilder);
             skipClrVisibilityChecks.SkipVisibilityChecksFor(viewType.GetTypeInfo());
 
             proxyTypeBuilder = proxyModuleBuilder.DefineType(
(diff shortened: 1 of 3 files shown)


AArnott:
CC: @davidwrighton

src/Microsoft.VisualStudio.Composition/Reflection/SkipClrVisibilityChecks.cs
```diff
@@ -64,33 +65,64 @@ internal SkipClrVisibilityChecks(AssemblyBuilder assemblyBuilder, ModuleBuilder
...
+            if (ReflectionHelpers.IsPublic(typeInfo.AsType()))
+            {
+                return ImmutableHashSet<AssemblyName>.Empty;
+            }
+            else
+            {
+                var result = ImmutableHashSet<AssemblyName>.Empty
+                    .Add(typeInfo.Assembly.GetName());
+
+                // If this type has a base type defined in another assembly that is also internal
```
AlexGhiondea: Did you mean to leave the commented code here?

src/Microsoft.VisualStudio.Composition/Configuration/MetadataViewGenerator.cs
```diff
@@ -79,36 +80,48 @@ internal static class MetadataViewGenerator
...
+            Requires.NotNull(viewType, nameof(viewType));
             Assumes.True(Monitor.IsEntered(MetadataViewFactories));
-            if (transparentProxyModuleBuilder == null)
+
+            // Dynamic assemblies are relatively expensive. We want to create as few as possible.
+            // For each unique set of skip visibility check assemblies, we need a new dynamic assembly
+            // because the CLR will not honor any additions to that set once the first generated type is closed.
+            // So maintain a dictionary to point at dynamic modules based on the set of skip visiblity check assemblies they were generated with.
+            var skipVisibilityCheckAssemblies = SkipClrVisibilityChecks.GetSkipVisibilityChecksRequirements(viewType);
+            if (!TransparentProxyModuleBuilderByVisibilityCheck.TryGetValue(skipVisibilityCheckAssemblies, out ModuleBuilder moduleBuilder))
```
AlexGhiondea: If the type is public, do you still need to do this? (i.e when `skipVisibilityCheckAssemblies` is ` ImmutableHashSet<AssemblyName>.Empty`

src/Microsoft.VisualStudio.Composition/Reflection/SkipClrVisibilityChecks.cs
AArnott: Yes, since it really feels like it *should* be here, and I want it to remind me to ask the CLR folks why I don't need visibility into transitive assemblies when I only ask to skip visibility checks for the first assembly.

src/Microsoft.VisualStudio.Composition/Configuration/MetadataViewGenerator.cs
AArnott: When the set is empty, we still need a dynamic assembly to generate into. Our lookup dictionary needs an entry for the empty set as well so we don't end up special casing empty sets -- we execute the same code regardless of the size of the set.

src/Microsoft.VisualStudio.Composition/Reflection/SkipClrVisibilityChecks.cs
davidwrighton: Because what's happening there is that access checks are only done on the metadata as written. For instance, A derives from B derives from C. As written, A only knows about B, so the access check from A is only to see if it can access B. For access to C, the only access check is that B is allowed to access C. These rules are straightforward as long as no techniques are used to skirt around access checks, but when you do, you can see surprises like this.

AArnott: OK, thanks. If A defines a member that my proxy implements, evidently that doesn't require C's visibility into A. I guess if A defines a member with a Type that is internal to A, at that point C couldn't implement it without an explicit skip visibility check for A. Yes?

PR #95, opened 2018-05-10
Re-enable MEF parts to omit the property setter on [ImportMany] properties

### LINKED ISSUE #94: `[ImportMany]` collection property has no setter, and is improperly rejected
The fix for #61 ended up regressing the scenario when an `ImportMany` collection has no setter but needn't have one because it is set by the importing constructor.

AArnott:
The case where this would actually work at runtime is when the MEF part's ImportingConstructor sets the value of the ImportMany property to an (empty) collection, in which case MEF will simply add elements to it and never have to access the setter.

Fixes #94

### DIFF
--- a/src/Microsoft.VisualStudio.Composition/Configuration/AttributedPartDiscovery.cs
+++ b/src/Microsoft.VisualStudio.Composition/Configuration/AttributedPartDiscovery.cs
@@ -312,7 +312,7 @@ namespace Microsoft.VisualStudio.Composition
 
             if (importAttribute != null)
             {
-                this.ThrowOnInvalidImportingMemberOrParameter(member);
+                this.ThrowOnInvalidImportingMemberOrParameter(member, isImportMany: false);
 
                 Type contractType = GetTypeIdentityFromImportingType(importingType, importMany: false);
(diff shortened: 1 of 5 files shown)


src/Microsoft.VisualStudio.Composition/Configuration/AttributedPartDiscovery.cs
```diff
@@ -333,7 +333,7 @@ private bool TryCreateImportDefinition(Type importingType, ICustomAttributeProvi
             }
             else if (importManyAttribute != null)
             {
-                this.ThrowOnInvalidImportingMemberOrParameter(member);
+                this.ThrowOnInvalidImportingMemberOrParameter(member, isImportMany: true);
```
dgriffen: I think it would make more sense just to omit the call in this case since setting `isImportMany` to true makes the function do nothing.

AArnott: The fact that it does nothing is an implementation detail of the method. In the future we may add more validations to importing members that may apply to ImportMany.

PR #109, opened 2019-01-15
Avoid torn state race condition in RuntimeImport.IsLazy

### LINKED ISSUE #108: random ArgumentNullException in VS MEF when using VS
It is a random problem reported by customers, and it has random stack in the product side (random place accessing MEF), but share the same stack similar to the one I pasted here.

After looking into the MEF code, I think it is a MEF code bug, independent to the composition.  The only explanation I can get is that the bug is inside RuntimeImport.IsLazy implementation.  It runs into a race condition, because Nullable<bool> is not thread safe.  When one thread computed the value and sets the value to true, in the split time, another thread can see the value has been set, and is false, which leads it to think valueFunction is null.

```
Microsoft.Requires.NotNull[T](T,String)
Microsoft.VisualStudio.Composition.RuntimeExportProviderFactory.RuntimeExportProvider.GetValueForImportElement(RuntimePartLifecycleTracker,RuntimeImport,RuntimeExport,Func`3)
Microsoft.VisualStudio.Composition.RuntimeExportProviderFactory.RuntimeExportProvider.GetValueForImportSite(RuntimePartLifecycleTracker,RuntimeImport)
Microsoft.VisualStudio.Composition.RuntimeExportProviderFactory.RuntimeExportProvider.RuntimePartLifecycleTracker.SatisfyImports()
Microsoft.VisualStudio.Composition.ExportProvider.PartLifecycleTracker.SatisfyImmediateImports()
Microsoft.VisualStudio.Composition.ExportProvider.PartLifecycleTracker.MoveNext(PartLifecycleState)
Microsoft.VisualStudio.Composition.ExportProvider.PartLifecycleTracker.MoveToState(PartLifecycleState)
Microsoft.VisualStudio.Composition.ExportProvider.PartLifecycleTracker.MoveToStateTransitively(PartLifecycleState,HashSet`1)
Microsoft.VisualStudio.Composition.ExportProvider.PartLifecycleTracker.MoveToStateTransitively(PartLifecycleState,HashSet`1)
Microsoft.VisualStudio.Composition.ExportProvider.PartLifecycleTracker.MoveToStateTransitively(PartLifecycleState,HashSet`1)
Microsoft.VisualStudio.Composition.ExportProvider.PartLifecycleTracker.MoveToStateTransitively(PartLifecycleState,HashSet`1)
Microsoft.VisualStudio.Composition.ExportProvider.PartLifecycleTracker.MoveToStateTransitively(PartLifecycleState,HashSet`1)
Microsoft.VisualStudio.Composition.ExportProvider.PartLifecycleTracker.MoveNext(PartLifecycleState)
Microsoft.VisualStudio.Composition.ExportProvider.PartLifecycleTracker.MoveToState(PartLifecycleState)
Microsoft.VisualStudio.Composition.ExportProvider.PartLifecycleTracker.GetValueReadyToExpose()
Microsoft.VisualStudio.Composition.RuntimeExportProviderFactory.RuntimeExportProvider.<>c__DisplayClass15_0.<GetExportedValueHelper>b__0()
Microsoft.VisualStudio.Composition.DelegateServices.<>c__DisplayClass2_0`1.<As>b__0()
System.Lazy`1.CreateValue()
System.Runtime.ExceptionServices.ExceptionDispatchInfo.Throw
```

AArnott:
There seems to be a race condition due to a non-atomic pair of memory writes while we lazily compute the value for `RuntimeImport.IsLazy`. 

A few hypotheses to explain the race condition:

1. Consider the case of the CPU reordering field write instructions, [which is allowed on some architectures, but not AMD64](https://en.wikipedia.org/wiki/Memory_ordering#In_symmetric_multiprocessing_(SMP)_microprocessor_systems). So that hypothesis is dubious.
1. The fields `hasValue` and `value` inside the `Nullable<T>` type are laid out in that order. So after a method constructs a new `Nullable<T>` value, it copies it to the field location, setting `hasValue` *first*. Before it copies the second field of the struct, another thread copies the entire struct to its own local stack, which now observes torn state of `hasValue=true;value=false` since `value` has not yet been set to true by the original thread. This seems to me the most plausible hypothesis.
1. Cache lines across cores: the `value=false` was cached by the reading thread and `hasValue` was not, such that `hasValue=true` is observed as the current value, but the stale `value=false` value is read from the cache. 

Any of these *should* be fixed by this change.

(hopefully) fixes #108

### DIFF
--- a/src/Microsoft.VisualStudio.Composition/NullableBool.cs
+++ b/src/Microsoft.VisualStudio.Composition/NullableBool.cs
@@ -1,0 +1,62 @@
+﻿// Copyright (c) Microsoft. All rights reserved.
+
+namespace Microsoft.VisualStudio.Composition
+{
+    using System;
+
+    /// <summary>
+    /// A nullable boolean value with atomic reads and writes.
+    /// </summary>
+    internal struct NullableBool
+    {
+        /// <summary>
+        /// A tri-state backing field for the <see cref="Value"/> property. 0 means not computed, -1 means false and 1 means true.
+        /// </summary>
+        /// <remarks>
+        /// We use a tri-state field to support lock-free atomic writes.
+        /// </remarks>
+        private int state;
+
+        /// <summary>
+        /// Initializes a new instance of the <see cref="NullableBool"/> struct.
+        /// </summary>
+        /// <param name="value">The initial value of the boolean.</param>
+        internal NullableBool(bool value)
+        {
+            this.state = value ? 1 : -1;
+        }
+
+        /// <summary>
+        /// Wraps a boolean value in a <see cref="NullableBool"/> struct.
+        /// </summary>
+        /// <param name="value">The boolean value to wrap.</param>
+        public static implicit operator NullableBool(bool value) => new NullableBool(value);
+
+        /// <summary>
+        /// Gets a value indicating whether the boolean value has been computed.
+        /// </summary>
+        internal bool HasValue => this.state != 0;
+
+        /// <summary>
+        /// Gets or sets a value indicating whether the boolean value is <c>true</c>.
+        /// </summary>
+        internal bool Value
+        {
+            get
+            {
+                switch (this.state)
+                {
+                    case 1: return true;
+                    case -1: return false;
+                    case 0: throw new InvalidOperationException("No value.");
+                    default: throw Assumes.NotReachable();
+                }
+            }
+
+            set
+            {
+                this.state = value ? 1 : -1;
+            }
+        }
+    }
+}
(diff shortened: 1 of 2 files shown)


src/Microsoft.VisualStudio.Composition/NullableBool.cs
```diff
@@ -0,0 +1,62 @@
+﻿// Copyright (c) Microsoft. All rights reserved.
+
+namespace Microsoft.VisualStudio.Composition
+{
+    using System;
+
+    /// <summary>
+    /// A nullable boolean value with atomic reads and writes.
+    /// </summary>
```
drewnoakes: Future readers may appreciate a remark such as: `Nullable<bool>` is larger than the machine's word size and is subject to tearing, which can result in invalid state during data races.

AArnott: Done.

PR #111, opened 2019-01-31
Avoid assembly loading during composition happy path

dgriffen:
Currently MEF composition loads a lot of assemblies unnecessarily. Ideally composition should not load any assemblies but that is a stretch goal. This PR fixes many cases of unnecessary assembly loading. Testing with VS2019 RC.3 reduced assembly loads during composition from 372 to 131 and composition time reduced by 70%


src/Microsoft.VisualStudio.Composition/Reflection/TypeRef.cs
```diff
@@ -242,7 +257,26 @@ public TypeRef MakeGenericTypeRef(ImmutableArray<TypeRef> genericTypeArguments)
 
             // We use the resolver parameter instead of the field here because this TypeRef instance
             // might have been constructed by TypeRef.Get(Type) and thus not have a resolver.
-            return new TypeRef(this.Resolver, this.AssemblyName, this.assemblyId, this.MetadataToken, this.FullName, this.TypeFlags, this.GenericTypeParameterCount, genericTypeArguments);
+            return new TypeRef(this.Resolver, this.AssemblyName, this.assemblyId, this.MetadataToken, this.FullName, this.TypeFlags, this.GenericTypeParameterCount, genericTypeArguments, this.ElementTypeRef);
+        }
+
+        public bool IsAssignableFrom(Type other)
```
AArnott: Is the goal to only avoid an assembly load on the *receiving* side? If you want to avoid loading *either* assembly, should you perhaps take `TypeRef` instead of `Type` for your parameter?

src/Microsoft.VisualStudio.Composition/Reflection/TypeRef.cs
```diff
@@ -242,7 +257,26 @@ public TypeRef MakeGenericTypeRef(ImmutableArray<TypeRef> genericTypeArguments)
...
+        {
+            if (other.EnumTypeBaseTypesAndInterfaces().Select(type => {
```
AArnott: We should probably avoid these delegates in these code paths to keep GC pressure down in a perf critical path. If you see it everywhere else nearby, then never mind. But I would expect we'd want to minimize here.

src/Microsoft.VisualStudio.Composition/RuntimeExportProviderFactory+RuntimeExportProvider.cs
```diff
@@ -100,7 +100,7 @@ private void ThrowIfExportedValueIsNotAssignableToImport(RuntimeComposition.Runt
 
                 if (exportedValue != null)
                 {
-                    if (!import.ImportingSiteTypeWithoutCollection.GetTypeInfo().IsAssignableFrom(exportedValue.GetType()))
+                    if (!import.ImportingSiteTypeWithoutCollectionRef.IsAssignableFrom(exportedValue.GetType()))
```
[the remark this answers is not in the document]
dgriffen: Oh whoops, I meant to remove that.

src/Microsoft.VisualStudio.Composition/Reflection/TypeRef.cs
dgriffen: The minimal goal was to only avoid the assembly load on the receiving side. From what I understood the general case for improvement would be a modified MEF cache, in that case won't most exporting types have to have their assembly loaded already to scan for new parts?

AArnott: Keep in mind that the assignability checks happen *during composition*, which applies to *all* assemblies, whether they were changed or not. So if you load the 'right side' of this test, then a *lot* of assemblies will load during composition because most assemblies export *something*, whether or not they were changed in the last cache invalidating update.

dgriffen: Ah, I thought the checks only happened on the boundary between modified and non-modified assemblies. In that case what I think will be necessary is recording the type hierarchy inside the typerefs.

dgriffen:
TypeRef work should be working now. The test that verifies if assemblies are not loaded is `ComposableAssembliesAssignabilityChecks`

src/Microsoft.VisualStudio.Composition/ReflectionHelpers.cs
```diff
@@ -106,22 +111,16 @@ internal static Assignability IsAssignableTo(ImportDefinitionBinding import, Exp
...
-                // Utilize the standard assignability checks for everything else.
-                if (receivingType.GetTypeInfo().IsAssignableFrom(exportingType.GetTypeInfo()))
-                {
-                    return Assignability.Definitely;
-                }
-
                 bool valueTypeKnownExactly =
                     export.ExportingMemberRef == null || // When [Export] appears on the type itself, we instantiate that exact type.
-                    exportingType.GetTypeInfo().IsSealed;
+                    exportingType.ResolvedType.GetTypeInfo().IsSealed;
```
AArnott: Of the remaining `ResolvedType` uses, what % of assemblies end up getting loaded after the changes you've already made? What % of assemblies were loaded before your changes?

src/Microsoft.VisualStudio.Composition/ReflectionHelpers.cs
```diff
@@ -677,6 +691,40 @@ internal static int GetMetadataTokenSafe(this MemberInfo memberInfo)
             return memberInfo.MetadataToken;
         }
 
+        internal static bool TryGetContainerType(Type type, out TypeInfo containerType)
+        {
+            Requires.NotNull(type, nameof(type));
+
+            // Discover the ICollection<T> or ICollection<Lazy<T, TMetadata>> interface implemented by this type.
+            var icollectionTypes =
```
AArnott: This is a non-trivial bit of code that you seem to have copied from `PartDiscovery.GetElementTypeFromMany`. Can you share the code instead? That way perf and functional changes will be consistent between the two.

src/Microsoft.VisualStudio.Composition/ReflectionHelpers.cs
```diff
@@ -677,6 +691,40 @@ internal static int GetMetadataTokenSafe(this MemberInfo memberInfo)
...
+                let genericTypeDef = ifaceInfo.GetGenericTypeDefinition()
+                where genericTypeDef.Equals(typeof(ICollection<>)) || genericTypeDef.Equals(typeof(IEnumerable<>)) || genericTypeDef.Equals(typeof(IList<>))
+                select ifaceInfo;
+
+            containerType = icollectionTypes.FirstOrDefault();
+
+            return containerType != null;
+        }
+
+        internal static Type GetElementType(Type type)
```
AArnott: This appears to be functionally equivalent to `PartDiscovery.GetElementTypeFromMany`. Can we share code instead of reimplementing it?

src/Microsoft.VisualStudio.Composition/RuntimeComposition.cs
```diff
@@ -209,16 +209,15 @@ private static RuntimeImport CreateRuntimeImport(ImportDefinitionBinding importD
             }
         }
 
-        private static RuntimeExport CreateRuntimeExport(ExportDefinition exportDefinition, Type partType, MemberRef exportingMemberRef, Resolver resolver)
+        private static RuntimeExport CreateRuntimeExport(ExportDefinition exportDefinition, TypeRef partType, MemberRef exportingMemberRef, Resolver resolver)
```
AArnott: nit: rename parameter to `partTypeRef`

src/Microsoft.VisualStudio.Composition/RuntimeComposition.cs
```diff
@@ -227,15 +226,15 @@ private static RuntimeExport CreateRuntimeExport(ExportDefinitionBinding exportD
...
             return CreateRuntimeExport(
                 exportDefinitionBinding.ExportDefinition,
-                partDefinitionType,
+                partDefinitionTypeRef,
                 exportDefinitionBinding.ExportingMemberRef,
                 resolver);
         }
 
-        [DebuggerDisplay("{" + nameof(RuntimePart.TypeRef) + "." + nameof(Reflection.TypeRef.ResolvedType) + ".FullName,nq}")]
+        [DebuggerDisplay("{" + nameof(RuntimePart.TypeRef) + ".FullName,nq}")]
```
AArnott: tip: use `nameof(TypeRef.FullName)` here because it's referencing a symbol we define ourselves and might change. We didn't do this before because it was *Type*.FullName, which will never change. But *TypeRef*.FullName might be removed/renamed in the future.

src/Microsoft.VisualStudio.Composition/RuntimeComposition.cs
```diff
@@ -498,7 +509,29 @@ public RuntimeExport(string contractName, TypeRef declaringTypeRef, MemberRef me
...
+                    {
+                        return this.exportedValueTypeRef;
+                    }
+                    else if (this.MemberRef == null)
+                    {
+                        return this.DeclaringTypeRef;
+                    }
+                    else
+                    {
+                        return TypeRef.Get(this.ExportedValueType, this.DeclaringTypeRef.Resolver);
```
AArnott: Similar here as an earlier comment: 
1. Should we cache this into the field for the next caller?
1. Can we invert this so that the TypeRef can be calculated without any assembly loads and the other non-ref property simply pulls on this property and resolves the result?

src/tests/Microsoft.VisualStudio.Composition.Tests/AssembliesLazyLoadedTests.cs
```diff
@@ -130,6 +130,31 @@ public async Task ComposableAssembliesLazyLoadedByLazyImport()
             }
         }
 
+        /// <summary>
+        /// Verifies that the assemblies that MEF parts belong to are only loaded when their parts are actually instantiated.
+        /// </summary>
+        [SkippableFact]
```
AArnott: In addition to these tests that you're adding, do tests pass if you merge [`dev/andarno/LazyCompose`](https://github.com/AArnott/vs-mef/commit/d224f51f5253c1fe02dbe69b36b81858370ce743) into your branch? If it passes, please do keep that merge and push it into your branch to lock in the higher test level.

src/Microsoft.VisualStudio.Composition/Configuration/SerializationContextBase.cs
```diff
@@ -688,7 +702,16 @@ protected void Write(IReadOnlyDictionary<string, object> metadata)
 
                 // Special case certain values to avoid defeating lazy load later.
                 // Check out the ReadMetadata below, how it wraps the return value.
-                var serializedMetadata = new LazyMetadataWrapper(metadata.ToImmutableDictionary(), LazyMetadataWrapper.Direction.ToSubstitutedValue, this.Resolver);
+                IReadOnlyDictionary<string, object> serializedMetadata;
+                if (metadata is LazyMetadataWrapper)
+                {
+                    serializedMetadata = LazyMetadataWrapper.TryUnwrap(metadata);
```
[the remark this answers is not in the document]
dgriffen: Yes, the issue is that if this is already a LazyMetaDataWrapper then we want the internal structure since it will contain representations of the metadata that don't require assembly loads. If its not a LazyMetadataWrapper then we need to wrap it to produce the serializable forms of the metadata. I'll add a comment explaining what's going on.

src/Microsoft.VisualStudio.Composition/ExportDefinitionBinding.cs
```diff
@@ -56,7 +56,17 @@ public bool IsStaticExport
...
-            get { return TypeRef.Get(this.ExportedValueType, this.PartDefinition.TypeRef.Resolver); }
+            get
+            {
+                if (this.ExportingMemberRef == null)
+                {
+                    return this.PartDefinition.TypeRef;
+                }
+                else
+                {
+                    return TypeRef.Get(this.ExportedValueType, this.PartDefinition.TypeRef.Resolver);
```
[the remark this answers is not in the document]
dgriffen: There was no particular reason to do so. I just wanted to keep the scope of this change to exported types and see what the performance improvement was before pursuing more aggressive optimizations.

src/Microsoft.VisualStudio.Composition/Configuration/SerializationContextBase.cs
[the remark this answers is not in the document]
AArnott: That depends. The LazyMetadataWrapper goes both directions. Unwrapping it may actually expose the types and lead to you serializing those instead of TypeRefs.

AArnott:
Should you remove the `[WIP]` header from the PR now that you're finalizing it?

src/Microsoft.VisualStudio.Composition/ReflectionHelpers.cs
```diff
@@ -81,22 +81,27 @@ internal static Assignability IsAssignableTo(ImportDefinitionBinding import, Exp
...
+                exportingType = exportingType.MakeGenericTypeRef(receivingType.GenericTypeArguments);
             }
 
-            if (typeof(Delegate).GetTypeInfo().IsAssignableFrom(receivingType.GetTypeInfo()) && typeof(Delegate).GetTypeInfo().IsAssignableFrom(exportingType.GetTypeInfo()))
+            // Quick assignability check to attempt to exit early if assignable, otherwise use the slower methods
+            if (receivingType.IsAssignableFrom(exportingType))
+            {
+                return Assignability.Definitely;
+            }
+            else if (typeof(Delegate).GetTypeInfo().IsAssignableFrom(exportingType.ResolvedType) && typeof(Delegate).GetTypeInfo().IsAssignableFrom(receivingType.ResolvedType))
```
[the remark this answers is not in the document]
dgriffen: I don't think the order change should affect things. If they're the exact same delegate type then the assignability check above should pass, which is fine. If they are slightly different but assignable delegate types then this check will still take care of them. I can't think of a situation where the assignability check would pass but the one presented here wouldn't.

src/Microsoft.VisualStudio.Composition/ExportDefinitionBinding.cs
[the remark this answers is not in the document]
dgriffen: I thought about it and ended up fixing this for exported properties and exported fields.

AArnott:
>         }

Can we revise this to be in terms of `ExportedValueTypeRef`? Not only will that reduce code and match the pattern used elsewhere, but it will force all code paths to use your new code which will help prove that it works in all cases.

---
Refers to: src/Microsoft.VisualStudio.Composition/ExportDefinitionBinding.cs:75 in 8ce62e3. [](commit_id = 8ce62e32144ece38a9a76b888d1072b17a3a0ceb, deletion_comment = False)

src/Microsoft.VisualStudio.Composition/ExportDefinitionBinding.cs
```diff
@@ -51,12 +51,22 @@ public MemberInfo ExportingMember
...
+                    return this.PartDefinition.TypeRef;
+                }
+                else
+                {
+                    return ReflectionHelpers.GetExportedValueTypeRef(this.PartDefinition.TypeRef, this.ExportingMemberRef);
+                }
+            }
         }
 
         public Type ExportedValueType
```
AArnott: >ExportedValueType [](start = 20, length = 17)

Can we revise this to be in terms of `ExportedValueTypeRef`? Not only will that reduce code and match the pattern used elsewhere, but it will force all code paths to use your new code which will help prove that it works in all cases.

src/Microsoft.VisualStudio.Composition/ReflectionHelpers.cs
```diff
@@ -219,6 +233,34 @@ internal static bool IsStatic(this MemberInfo exportingMember)
             throw new NotSupportedException();
         }
 
+        internal static bool IsStatic(this MemberRef exportingMemberRef)
```
AArnott: >IsStatic(this MemberRef exportingMemberRef) [](start = 29, length = 43)

Why isn't this a property on `MemberRef` itself? Which derived type are you *not* checking in this method? It seems that every member might be static (except constructors, but there isn't a ConstructorRef type, I think). If you promote the property to the base type, you can remove this method.

src/Microsoft.VisualStudio.Composition/ReflectionHelpers.cs
```diff
@@ -537,6 +594,27 @@ internal static Type GetExportedValueType(Type declaringType, MemberInfo exporti
...
+
+            if (exportingMemberRef is FieldRef || exportingMemberRef is PropertyRef)
+            {
+                return ReflectionHelpers.GetMemberTypeRef(exportingMemberRef);
+            }
+
+            var exportingMethodRef = exportingMemberRef.MemberInfo as MethodInfo;
+            if (exportingMethodRef != null)
+            {
+                return TypeRef.Get(GetContractTypeForDelegate(exportingMethodRef), declaringTypeRef.Resolver);
```
AArnott: >declaringTypeRef [](start = 83, length = 16)

`declaringTypeRef` is dereferenced without ever checking for a null value.

src/Microsoft.VisualStudio.Composition/RuntimeComposition.cs
AArnott: The way you've fixed this, you've removed "TypeRef." from the string. Only the last member (FullName) contributes to the string.

---
In reply to: [262675122](https://github.com/Microsoft/vs-mef/pull/111#discussion_r262675122) [](ancestors = 262675122)

src/Microsoft.VisualStudio.Composition/Reflection/ResolverExtensions.cs
```diff
@@ -162,19 +162,15 @@ internal static void GetInputAssemblies(this TypeRef typeRef, ISet<AssemblyName>
...
-                    assemblies.Add(baseType.GetTypeInfo().Assembly.GetName());
-                }
-
-                // Interfaces may also define [InheritedExport] attributes, metadata view filters, etc.
-                foreach (var iface in type.GetTypeInfo().GetInterfaces())
+                //var type = typeRef.Resolve();
+                foreach (var baseType in typeRef.BaseTypes)
                 {
-                    assemblies.Add(iface.GetTypeInfo().Assembly.GetName());
+                    assemblies.Add(baseType.ResolvedType.GetTypeInfo().Assembly.GetName());
```
AArnott: >assemblies.Add(baseType.ResolvedType.GetTypeInfo().Assembly.GetName()) [](start = 20, length = 70)

Why is resolving the TypeRef necessary, consider TypeRef has a property that stores the assembly name?

src/Microsoft.VisualStudio.Composition/Reflection/ResolverExtensions.cs
```diff
@@ -162,19 +162,15 @@ internal static void GetInputAssemblies(this TypeRef typeRef, ISet<AssemblyName>
...
                 // Resolving a TypeRef is a necessary cost in order to identify the transitive closure of base types.
-                var type = typeRef.Resolve();
-                foreach (var baseType in type.EnumTypeAndBaseTypes())
-                {
-                    assemblies.Add(baseType.GetTypeInfo().Assembly.GetName());
-                }
-
-                // Interfaces may also define [InheritedExport] attributes, metadata view filters, etc.
-                foreach (var iface in type.GetTypeInfo().GetInterfaces())
+                //var type = typeRef.Resolve();
```
AArnott: > //var type = typeRef.Resolve(); [](start = 15, length = 32)

Do we need to keep this commented line?
Also the comment above it seems to be obsolete.

src/Microsoft.VisualStudio.Composition/Reflection/ResolverExtensions.cs
```diff
@@ -162,19 +162,15 @@ internal static void GetInputAssemblies(this TypeRef typeRef, ISet<AssemblyName>
...
+                foreach (var baseType in typeRef.BaseTypes)
```
AArnott: >BaseTypes [](start = 49, length = 9)

This includes interfaces, which we weren't including before. That puts our cache in a more likely position to be considered stale when assemblies change that only contribute interfaces. Can we filter out the interfaces in this enumeration?

dgriffen: The previous code enumerated both in separate loops. The end effect here should be the same.

src/tests/Microsoft.VisualStudio.Composition.Tests/AssembliesLazyLoadedTests.cs
dgriffen: They don't pass since there are still some cases where assemblies will get loaded. I'll leave it unmerged for now since I think fully fixing that is out of scope for this PR.

src/Microsoft.VisualStudio.Composition/LazyMetadataWrapper.cs
```diff
@@ -195,6 +197,32 @@ public bool Remove(KeyValuePair<string, object> item)
...
+        internal static bool TryGetMetadataValueTypeRef(IReadOnlyDictionary<string, object> metadata, string key, Resolver resolver, out object typeRef)
+        {
+            typeRef = null;
+
+            if (metadata is LazyMetadataWrapper lazyMetadata)
+            {
+                metadata = lazyMetadata.underlyingMetadata;
+            }
+
+            if (!metadata.TryGetValue(key, out var value))
```
AArnott: >metadata [](start = 17, length = 8)

You're dereferencing this argument without validating that it isn't null coming in from the caller.

src/Microsoft.VisualStudio.Composition/LazyMetadataWrapper.cs
```diff
@@ -195,6 +197,32 @@ public bool Remove(KeyValuePair<string, object> item)
...
 
+        internal static bool TryGetMetadataValueTypeRef(IReadOnlyDictionary<string, object> metadata, string key, Resolver resolver, out object typeRef)
+        {
+            typeRef = null;
+
+            if (metadata is LazyMetadataWrapper lazyMetadata)
+            {
+                metadata = lazyMetadata.underlyingMetadata;
+            }
+
```
AArnott: I'm still a bit puzzled as to why you're unwrapping this without regard to the direction the wrapper was going.

src/Microsoft.VisualStudio.Composition/LazyMetadataWrapper.cs
```diff
@@ -195,6 +197,32 @@ public bool Remove(KeyValuePair<string, object> item)
             return metadata;
         }
 
+        internal static bool TryGetMetadataValueTypeRef(IReadOnlyDictionary<string, object> metadata, string key, Resolver resolver, out object typeRef)
```
AArnott: >object [](start = 137, length = 6)

Why is `typeRef` typed as `object` here? If it's always a `TypeRef`, why not type it as such? Or if it might not be one, why name it that way?
From the assignment below, it looks like in fact this value might be set to whatever metadata value the metadata is set to, including `Type` itself.
At a minimum, let's please add an xml doc comment explaining what this method and all its parameters and return values signify.

src/Microsoft.VisualStudio.Composition/PartDiscovery.cs
```diff
@@ -202,13 +202,46 @@ protected internal static Type GetTypeIdentityFromImportingType(Type type, bool
...
+            {
+                throw new ArgumentException(string.Format(Strings.ImportManyOnNonCollectionType, type.FullName), nameof(type));
+            }
+        }
+
+        internal static bool TryGetElementTypeFromMany(Type type, out Type elementType)
+        {
+            Requires.NotNull(type, nameof(type));
+            elementType = null;
+
```
AArnott: nit: move this to just above the `return false;` line. This will force anyone who returns `true` to remember to set this out parameter, whereas right now they could get away with forgetting and send null to the caller.

src/Microsoft.VisualStudio.Composition/Reflection/PropertyRef.cs
```diff
@@ -32,25 +32,31 @@ public class PropertyRef : MemberRef, IEquatable<PropertyRef>
...
             this.Name = name;
+            this.PropertyTypeRef = propertyTypeRef;
         }
 
         public PropertyRef(PropertyInfo propertyInfo, Resolver resolver)
             : base(propertyInfo, resolver)
         {
+            this.getMethodMetadataToken = propertyInfo?.GetMethod?.MetadataToken;
+            this.setMethodMetadataToken = propertyInfo?.SetMethod?.MetadataToken;
             this.Name = propertyInfo.Name;
```
AArnott: I'm curious whether this was just an elective change or if there is some significance behind it that I can't see.

src/Microsoft.VisualStudio.Composition/Reflection/TypeRef.cs
```diff
@@ -106,6 +117,15 @@ private TypeRef(Resolver resolver, Type type)
 
         public ImmutableArray<TypeRef> GenericTypeArguments { get; private set; }
 
+        /// <summary>
+        /// Gets the full list of base types and interfaces for this instance.
+        /// This list will only be populated if this instance was created with shallow set to false.
```
AArnott: >This list will only be populated if this instance was created with shallow set to false. [](start = 12, length = 88)

What will it be if `shallow` is true? Will it be empty or "uninitialized" (as only an `ImmutableArray<T>` can be). How can I tell if this is the case? Is there a `IsShallow` property or something I can check?

src/Microsoft.VisualStudio.Composition/Reflection/TypeRef.cs
```diff
@@ -242,7 +262,38 @@ public TypeRef MakeGenericTypeRef(ImmutableArray<TypeRef> genericTypeArguments)
...
-            return new TypeRef(this.Resolver, this.AssemblyName, this.assemblyId, this.MetadataToken, this.FullName, this.TypeFlags, this.GenericTypeParameterCount, genericTypeArguments);
+            return new TypeRef(this.Resolver, this.AssemblyName, this.assemblyId, this.MetadataToken, this.FullName, this.TypeFlags, this.GenericTypeParameterCount, genericTypeArguments, this.BaseTypes, this.ElementTypeRef);
+        }
+
+        /// <summary>
+        /// Checks if the type represented by the given TypeRef can be assigned to the type represented by this instance.
+        /// The assignability check is done by traversing all the base types and interfaces of the given TypeRef to check
+        /// if any of them are equal to this instance. Should that fail, the CLR is asked to check for assignability
+        /// which will trigger an assembly load.
+        /// </summary>
```
AArnott: nit: move this to a `<remarks>` section. Since in effect the result is reliable, it's not super important to the caller of this method to know how it's done.

src/Microsoft.VisualStudio.Composition/Reflection/TypeRef.cs
```diff
@@ -273,7 +346,8 @@ public bool Equals(TypeRef other)
                 && AssemblyNameComparer.Equals(this.AssemblyName, other.AssemblyName)
                 && this.IsArray == other.IsArray
                 && this.GenericTypeParameterCount == other.GenericTypeParameterCount
-                && this.GenericTypeArguments.EqualsByValue(other.GenericTypeArguments);
+                && this.GenericTypeArguments.EqualsByValue(other.GenericTypeArguments)
+                && this.BaseTypes.EqualsByValue(other.BaseTypes);
             return result;
```
AArnott: Can you remind me why we need to check this? This strikes me as an expensive comparison, and as we've already confirmed by this point that the metadata tokens and generic type arguments are the same, why do we need to verify the base type hierarchy hasn't changed? If it's to distinguish between shallow and not-shallow instances, can't we just compare `this.BaseTypes.IsEmpty == other.BaseTypes.IsEmpty` instead as that would be super-quick?

dgriffen: This was to distinguish shallow types from non-shallow types for the purposes of serialization. I can change this to just use `IsEmpty`.

src/Microsoft.VisualStudio.Composition/LazyMetadataWrapper.cs
dgriffen: I'll add the check. The reason it hasn't been relevant yet is because the only time we construct a `LazyMetadataWrapper` in the `ToSubstitutedValue` direction is during serialization.

src/Microsoft.VisualStudio.Composition/Reflection/PropertyRef.cs
dgriffen: It stops assembly loading in scenarios where either the get method or set method metadata is null. Right now if the metadata is intentionally null we will load the assembly this property is defined it just to make sure.

src/Microsoft.VisualStudio.Composition/Reflection/TypeRef.cs
dgriffen: There is currently no way to determine if a `TypeRef` is shallow. Generally speaking the `TypeRef`s you get ahold of will not be shallow since they will be at the bottom of the type hierarchy.

src/tests/Microsoft.VisualStudio.Composition.Tests/AssembliesLazyLoadedTests.cs
```diff
@@ -324,6 +395,11 @@ internal void TestExternalExportWithLazy(string lazyLoadedAssemblyPath)
                 AssertEx.True(GetLoadedAssemblies().Any(a => a.Location.Equals(lazyLoadedAssemblyPath, StringComparison.OrdinalIgnoreCase)));
             }
 
+            internal void TestComposition(params string[] assemblyPaths)
```
AArnott: >TestComposition [](start = 26, length = 15)

Can we give this a more descriptive name? e.g. `AssertAssembliesNotLoaded`

src/Microsoft.VisualStudio.Composition/Reflection/TypeRef.cs
AArnott: Do non-shallow objects *ever* have an *empty* array here? If not, then an empty array could be an indicator and could be documented (or used to implement a property getter: `internal bool IsShallow => this.BaseTypes.IsEmpty`.) Do value types and reference types both have base types that are added to this array?

---
In reply to: [266672285](https://github.com/Microsoft/vs-mef/pull/111#discussion_r266672285) [](ancestors = 266672285)

src/Microsoft.VisualStudio.Composition/LazyMetadataWrapper.cs
```diff
@@ -195,6 +197,32 @@ public bool Remove(KeyValuePair<string, object> item)
...
+            }
+
+            if (value is ISubstitutedValue substitutedValue)
+            {
+                typeRef = substitutedValue.SubstitutedValueTypeRef;
+            }
+            else
+            {
+                typeRef = value;
+            }
```
AArnott: This could be an instance of `Type`, but it seems we should coerce it to be `TypeRef` since the caller has optimizations that only apply if it is a `TypeRef`.

src/Microsoft.VisualStudio.Composition/ReflectionHelpers.cs
AArnott: Now that you've promoted the `IsStatic` property to the base type, we can remove this method, right?

---
In reply to: [264505835](https://github.com/Microsoft/vs-mef/pull/111#discussion_r264505835) [](ancestors = 264505835)

src/Microsoft.VisualStudio.Composition/Reflection/TypeRef.cs
dgriffen: Non shallow objects can have an empty array. `object` is one example. This can also happen with TypeRefs to interfaces.

dgriffen:
Build error seems to be spurious. I don't have permission to re-queue.

AArnott:
I've requeued the build. But @dgriffen you might want to rebase your commits to squash at least some of these. :)

PR #112, opened 2019-02-06
Invalid runtime composition exception handling.

tinaschrepfer:
Add more information to metadata view exception.  Catch invalid runtime composition and throw exception.

### DIFF
--- a/src/Microsoft.VisualStudio.Composition/RuntimeComposition.cs
+++ b/src/Microsoft.VisualStudio.Composition/RuntimeComposition.cs
@@ -27,10 +27,20 @@ namespace Microsoft.VisualStudio.Composition
             Requires.NotNull(metadataViewsAndProviders, nameof(metadataViewsAndProviders));
             Requires.NotNull(resolver, nameof(resolver));
 
-            if (!parts.Any() || !metadataViewsAndProviders.Any())
+            var invalidArgExceptionMsg = "Invalid arguments passed to generate runtime composition.";
+            var invalidArgExceptionData = "InvalidCompositionException";
+            if (!parts.Any())
             {
-                var exception = new ArgumentException("Invalid arguments passed to generate runtime composition.");
-                exception.Data.Add("InvalidCompositionException", true);
+                var exception = new ArgumentException(invalidArgExceptionMsg, nameof(parts));
+                exception.Data.Add(invalidArgExceptionData, true);
+
+                throw exception;
+            }
+
+            if (!metadataViewsAndProviders.Any())
+            {
+                var exception = new ArgumentException(invalidArgExceptionMsg, nameof(metadataViewsAndProviders));
+                exception.Data.Add(invalidArgExceptionData, true);
 
                 throw exception;
             }
--- a/src/Microsoft.VisualStudio.Composition/RuntimeComposition.cs
+++ b/src/Microsoft.VisualStudio.Composition/RuntimeComposition.cs
@@ -21,11 +21,21 @@ namespace Microsoft.VisualStudio.Composition
         private readonly IReadOnlyDictionary<string, IReadOnlyCollection<RuntimeExport>> exportsByContractName;
         private readonly IReadOnlyDictionary<TypeRef, RuntimeExport> metadataViewsAndProviders;
 
+        public const string InvalidCompositionException = nameof(InvalidCompositionException);
+
         private RuntimeComposition(IEnumerable<RuntimePart> parts, IReadOnlyDictionary<TypeRef, RuntimeExport> metadataViewsAndProviders, Resolver resolver)
         {
             Requires.NotNull(parts, nameof(parts));
             Requires.NotNull(metadataViewsAndProviders, nameof(metadataViewsAndProviders));
             Requires.NotNull(resolver, nameof(resolver));
+
+            if (!parts.Any() || !metadataViewsAndProviders.Any())
+            {
+                var exception = new ArgumentException("Invalid arguments passed to generate runtime composition.");
+                exception.Data.Add(InvalidCompositionException, true);
+
+                throw exception;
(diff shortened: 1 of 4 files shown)


src/Microsoft.VisualStudio.Composition/RuntimeComposition.cs
```diff
@@ -21,12 +21,22 @@ public class RuntimeComposition : IEquatable<RuntimeComposition>
...
 
+        public const string InvalidCompositionException = nameof(InvalidCompositionException);
+
         private RuntimeComposition(IEnumerable<RuntimePart> parts, IReadOnlyDictionary<TypeRef, RuntimeExport> metadataViewsAndProviders, Resolver resolver)
         {
             Requires.NotNull(parts, nameof(parts));
             Requires.NotNull(metadataViewsAndProviders, nameof(metadataViewsAndProviders));
             Requires.NotNull(resolver, nameof(resolver));
 
+            if (!parts.Any() || !metadataViewsAndProviders.Any())
```
AArnott: Do we already have a unit test to verify that an empty `ComposableCatalog` can still compose and turn into an `ExportProvider` that doesn't throw here? If not, please add one.

src/Microsoft.VisualStudio.Composition/RuntimeComposition.cs
```diff
@@ -21,12 +21,22 @@ public class RuntimeComposition : IEquatable<RuntimeComposition>
         private readonly IReadOnlyDictionary<string, IReadOnlyCollection<RuntimeExport>> exportsByContractName;
         private readonly IReadOnlyDictionary<TypeRef, RuntimeExport> metadataViewsAndProviders;
 
+        public const string InvalidCompositionException = nameof(InvalidCompositionException);
```
AArnott: This seems like an odd public API to add. That sounds like an exception type, yet it's just a constant named after itself.

Can we keep it private?

src/Microsoft.VisualStudio.Composition/RuntimeComposition.cs
```diff
@@ -21,12 +21,22 @@ public class RuntimeComposition : IEquatable<RuntimeComposition>
...
+
         private RuntimeComposition(IEnumerable<RuntimePart> parts, IReadOnlyDictionary<TypeRef, RuntimeExport> metadataViewsAndProviders, Resolver resolver)
         {
             Requires.NotNull(parts, nameof(parts));
             Requires.NotNull(metadataViewsAndProviders, nameof(metadataViewsAndProviders));
             Requires.NotNull(resolver, nameof(resolver));
 
+            if (!parts.Any() || !metadataViewsAndProviders.Any())
+            {
+                var exception = new ArgumentException("Invalid arguments passed to generate runtime composition.");
```
AArnott: Can we separate the `parts.Any()` and `metadataViewAndProviders.Any()` checks and then throw an exception that actually calls out which argument is invalid? When users hit this (and possibly when _we_ hit this) it helps diagnosis when we know what exactly is wrong.

src/Microsoft.VisualStudio.Composition/RuntimeComposition.cs
AArnott: I see you've already done this. :)

src/tests/Microsoft.VisualStudio.Composition.Tests/RuntimeCompositionTests.cs
```diff
@@ -0,0 +1,17 @@
...
+
+    public class RuntimeCompositionTests
+    {
+        [Fact]
+        public void TestEmptyCatalogTest()
+        {
+            var configuration = CompositionConfiguration.Create(TestUtilities.EmptyCatalog);
+            var composition = RuntimeComposition.CreateRuntimeComposition(configuration);
+            var exports = composition.GetExports(typeof(IDisposable).ToString());
+            Assert.Equal(0, exports.Count);
```
AArnott: nit: `Assert.Empty(exports);` can result in a more useful error message when the test fails. Not a big deal, but something to consider next time.

src/tests/Microsoft.VisualStudio.Composition.Tests/RuntimeCompositionTests.cs
```diff
@@ -0,0 +1,17 @@
...
+    using Xunit;
+
+    public class RuntimeCompositionTests
+    {
+        [Fact]
+        public void TestEmptyCatalogTest()
+        {
+            var configuration = CompositionConfiguration.Create(TestUtilities.EmptyCatalog);
+            var composition = RuntimeComposition.CreateRuntimeComposition(configuration);
+            var exports = composition.GetExports(typeof(IDisposable).ToString());
```
AArnott: Using `.ToString()` seems odd here, as passing in an explicit contract name doesn't resemble the common use case in VS-MEF. Why not use `GetExports<IDisposable>()`?
But I think it serves the purpose either way. I just tend to prefer unit tests that look like the most common use case possible while still testing the intended thing. That way, if it fails, we can readily see how impactful the failure might be.

src/tests/Microsoft.VisualStudio.Composition.Tests/RuntimeCompositionTests.cs
```diff
@@ -0,0 +1,19 @@
+﻿namespace Microsoft.VisualStudio.Composition.Tests
+{
+    using System;
+    using Xunit;
+
+    public class RuntimeCompositionTests
```
AArnott: Can you also add a test for the 2 `ArgumentException` cases that you now throw?

tinaschrepfer: Done.

PR #120, opened 2019-03-19
Fix code coverage

AArnott:
Also remove obsolete uses of the Hosted variable


dgriffen:
Something seems to have gone weird with the codecov upload. I don't see the results anywhere.

AArnott:
Ya, this PR doesn't actually fix it ... yet.

PR #123, opened 2019-03-25
switch to coverlet for code coverage

dgriffen:


AArnott:
Can we rebase this fix on v16.0 though, so we get it working on those branches as well?

dgriffen:
@AArnott I rebased to v16.0 and the checks passed, can you merge this in?

PR #127, opened 2019-04-08
Handle overridden members of generic base classes

### LINKED ISSUE #125: Slow reflection path can fail with overidden properties
`Type.GetProperty` can throw an `AmbiguousMatchException` if a property overrides a property in a base type. This can cause issues in the slow reflection path of the [`PropertyRef.Resolve()`](https://github.com/Microsoft/vs-mef/blob/03b124efc4a6b8ae4f82962ba13ec8fa45aaa231/src/Microsoft.VisualStudio.Composition/Reflection/ResolverExtensions.cs#L56) method.

dgriffen:
Closes #125

### DIFF
--- a/src/Microsoft.VisualStudio.Composition/Reflection/ResolverExtensions.cs
+++ b/src/Microsoft.VisualStudio.Composition/Reflection/ResolverExtensions.cs
@@ -13,7 +13,7 @@ namespace Microsoft.VisualStudio.Composition.Reflection
 
     public static class ResolverExtensions
     {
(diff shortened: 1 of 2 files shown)


AArnott:
You're targeting `master` (which is post-16.0). Is that appropriate or do you want to service 16.0 with this change?

dgriffen:
> 
> 
> You're targeting `master` (which is post-16.0). Is that appropriate or do you want to service 16.0 with this change?

I think its fine to take this in 16.1/16.2. The fix in VS should stop the scenario we're seeing.

PR #128, opened 2019-04-10
Compare assembly identity instead of name

### LINKED ISSUE #126: TypeRef.Equals should check assembly MVID
Currently `TypeRef.Equals` does not check the MVID of its assembly, just the full assembly name. This is inconsistent with the check that `TryUseFastReflection` uses. Meaning that cache equality comparisions can return `true` but slow reflection would still have to be used.

dgriffen:
Closes #126

### DIFF
--- a/src/tests/Microsoft.VisualStudio.Composition.Tests/Reflection/TypeRefTests.cs
+++ b/src/tests/Microsoft.VisualStudio.Composition.Tests/Reflection/TypeRefTests.cs
@@ -26,7 +26,7 @@ namespace Microsoft.VisualStudio.Composition.Tests.Reflection
             const string assemblyNameFormat = "MyAssembly, Version={0}, Culture=neutral, PublicKeyToken=abcdef1234567890, processorArchitecture=MSIL";
             string assemblyNameV1 = string.Format(assemblyNameFormat, "1.0.0.0");
             string assemblyNameV2 = string.Format(assemblyNameFormat, "2.0.0.0");
-            this.TestAssemblyNameEqualityNotEqual(assemblyNameV1, assemblyNameV2, @"C:\MyAssembly.dll", @"C:\MyAssembly.dll");
+            this.TestAssemblyNameEqualityNotEqual(assemblyNameV1, assemblyNameV2, @"C:\MyAssembly.dll", @"C:\MyAssembly.dll", Guid.Empty, Guid.Empty);
         }
 
         [Fact]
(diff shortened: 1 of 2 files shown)


src/tests/Microsoft.VisualStudio.Composition.Tests/Reflection/TypeRefTests.cs
```diff
@@ -49,8 +49,10 @@ private void TestAssemblyNameEqualityNotEqual(string assemblyNameV1String, strin
             assemblyNameV2.CodeBase = codeBaseV2;
 #endif
 
-            TypeRef typeRefV1 = TypeRef.Get(TestUtilities.Resolver, assemblyNameV1, 0x02000001, "SomeType", TypeRefFlags.None, 0, ImmutableArray<TypeRef>.Empty, false, ImmutableArray<TypeRef>.Empty, null);
-            TypeRef typeRefV2 = TypeRef.Get(TestUtilities.Resolver, assemblyNameV2, 0x02000001, "SomeType", TypeRefFlags.None, 0, ImmutableArray<TypeRef>.Empty, false, ImmutableArray<TypeRef>.Empty, null);
+            StrongAssemblyIdentity assemblyIdentityV1 = new StrongAssemblyIdentity(assemblyNameV1, Guid.NewGuid());
+            StrongAssemblyIdentity assemblyIdentityV2 = new StrongAssemblyIdentity(assemblyNameV2, Guid.NewGuid());
+            TypeRef typeRefV1 = TypeRef.Get(TestUtilities.Resolver, assemblyIdentityV1, 0x02000001, "SomeType", TypeRefFlags.None, 0, ImmutableArray<TypeRef>.Empty, false, ImmutableArray<TypeRef>.Empty, null);
+            TypeRef typeRefV2 = TypeRef.Get(TestUtilities.Resolver, assemblyIdentityV2, 0x02000001, "SomeType", TypeRefFlags.None, 0, ImmutableArray<TypeRef>.Empty, false, ImmutableArray<TypeRef>.Empty, null);
```
AArnott: Can you add the new assertion without changing the existing ones? I feel like it's still valid to test equality after using TypeRef with just an assembly name, which we lose with your current iteration.
Secondly, your updated assertion is testing inequality when *both* the assembly name *and* the MVID are different, which I suspect would "pass" even without your product change, since we were already including assembly names. So what I'd like to see is the new assertion test with the *same* assembly name but a unique MVID, to narrow down what we're testing and better represent the bug you found.

dgriffen: The reason I had to change the existing assertions was that an assembly identity cannot be loaded in these scenarios since the dll does not exist on disk. I do think its a good idea to test changing either one individually though.

AArnott: Ah, yes ok I agree.

PR #129, opened 2019-05-09
Log assembly loads during build-time composition

AArnott:


src/Microsoft.VisualStudio.Composition.AppHost/CreateComposition.cs
```diff
@@ -44,6 +44,16 @@ public class CreateComposition : AppDomainIsolatedTask, ICancelableTask
...
 
+        /// <summary>
+        /// Gets the assemblies that have been logged as to their locations already.
+        /// </summary>
+        private readonly HashSet<Assembly> loggedAssemblies = new HashSet<Assembly>();
+
+        /// <summary>
+        /// Gets the names of assemblies that could not be found and have been logged.
+        /// </summary>
+        private readonly HashSet<string> loggedMissingAssemblyNames = new HashSet<string>(StringComparer.OrdinalIgnoreCase);
```
ikeras: Why do we ignore case here, but not for loggedAssemblies above?

src/Microsoft.VisualStudio.Composition.AppHost/CreateComposition.cs
```diff
@@ -44,6 +44,16 @@ public class CreateComposition : AppDomainIsolatedTask, ICancelableTask
         /// </summary>
         private readonly List<string> catalogAssemblyPaths = new List<string>();
 
+        /// <summary>
+        /// Gets the assemblies that have been logged as to their locations already.
```
ikeras: This might be clearer if you said "Gets the assemblies that have been found and logged."

src/Microsoft.VisualStudio.Composition.AppHost/CreateComposition.cs
```diff
@@ -380,13 +401,15 @@ private static bool Probe(string filePath, Version minimumVersion, out Assembly
...
                     {
                         assembly = Assembly.Load(name);
+                        this.LogAssemblyLoad(assembly, filePath);
                         return true;
                     }
                     catch (BadImageFormatException)
                     {
                         // This happens for some reference assemblies that can only be loaded in a reflection-only context.
                         // But we're not allowed in an assembly resolve event to return reflection-only loaded assemblies.
                         // So just return false to communicate the failure.
```
ikeras: You might modify the comment here slightly to say "So just log it and return false ..."

src/Microsoft.VisualStudio.Composition.AppHost/CreateComposition.cs
AArnott: Because we're storing strings in this set, whereas the *found* assemblies we're storing `Assembly` objects.

src/Microsoft.VisualStudio.Composition.AppHost/CreateComposition.cs
AArnott: Done

src/Microsoft.VisualStudio.Composition.AppHost/CreateComposition.cs
AArnott: Done

PR #131, opened 2019-05-15
Better handling of properties with default value attribute

ppease:
# Justification
VS MEF does not handle properties with the default value attribute on metadata views the same way that V1 of MEF handles them. Specifically if the metadata view is tagged with the attribute specifying a metadataview implementation, v1 of MEF ignores any export metadata associated with properties that have a default value attribute.

Our team really wants to update to VS MEF, but we have some unique scenarios where a lot of our code uses this pattern of setting the export metadata to something that is a different data type than the property itself, and we interpret that to set the property correctly in the constructor of our metadata view implementation. Currently in VS MEF this fails. The problem is the check in ImportMetadataViewConstraint.IsSatisfiedBy. In this check the constraint determines there is export metadata, and it makes sure it can be assigned to the property.

# Implementation
It seemed straight forward to just not create a MetadatumRequirement in the scenario that we had a metadataview implementation and the default value attribute was defined.  In all the testing I did V1 of MEF just completely ignored export metadata for these kinds of properties.

# Testing
I added a new property and new export data in the MetadataViewImplementationTests test class that cover this scenario.

### DIFF
--- a/src/tests/Microsoft.VisualStudio.Composition.Tests/MetadataViewImplementationTests.cs
+++ b/src/tests/Microsoft.VisualStudio.Composition.Tests/MetadataViewImplementationTests.cs
@@ -20,15 +20,17 @@ namespace Microsoft.VisualStudio.Composition.Tests
             Assert.IsType<MetadataViewClass>(export.Metadata);
             Assert.Equal("1", export.Metadata.A);
             Assert.Null(export.Metadata.B);
+            Assert.Equal(10, export.Metadata.PropertyWithMetadataThatDoesNotMatchType);
         }
 
-        [MefFact(CompositionEngines.V1)]
+        [MefFact(CompositionEngines.V1Compat)]
         public void MetadataViewImplementationViaImport(IContainer container)
         {
             var importingPart = container.GetExportedValue<ImportingPart>();
             Assert.IsType<MetadataViewClass>(importingPart.ImportingProperty.Metadata);
             Assert.Equal("1", importingPart.ImportingProperty.Metadata.A);
             Assert.Null(importingPart.ImportingProperty.Metadata.B);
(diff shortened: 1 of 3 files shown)


src/tests/Microsoft.VisualStudio.Composition.Tests/MetadataViewImplementationTests.cs
```diff
@@ -20,15 +20,17 @@ public void MetadataViewImplementationDirectQuery(IContainer container)
             Assert.IsType<MetadataViewClass>(export.Metadata);
             Assert.Equal("1", export.Metadata.A);
             Assert.Null(export.Metadata.B);
+            Assert.Equal(10, export.Metadata.PropertyWithMetadataThatDoesNotMatchType);
         }
 
-        [MefFact(CompositionEngines.V1)]
+        [MefFact(CompositionEngines.V1Compat)]
```
ppease: I didn't understand why this wasn't tagged as V1Compat.  I was assuming this would need to be tested with V1 of MEF as well as VS MEF so I switched it.

AArnott:
Sorry it has taken a while to respond. We like the PR, but are a bit concerned how it may affect Visual Studio itself. I hope to measure this in the next few weeks. Assuming we can mitigate whatever behavioral changes this makes, we'd like to accept this PR.

AArnott:
@ppease This is available on [our prerelease feed](https://dev.azure.com/azure-public/vside/_packaging?_a=package&feed=vssdk&package=Microsoft.VisualStudio.Composition&protocolType=NuGet&version=16.4.4-alpha) as 16.4.4-alpha. Will that unblock you?

ppease:
Yes this helps us.  Thanks for taking time to get this in!  It is so much nicer to be able to just depend on the latest nuget rather than having to maintain our own forked copy.

PR #146, opened 2019-07-18
Remove older Target Frameworks

AArnott:
Microsoft's long-term support for .NET Core 1.x and 2.0 [are over][LTS]. Customers targeting these versions are in a really bad place due to no security patches being offered and thus there are likely few to none out there.

Similarly, such a critical mass of customers already have net472 on their machines that many apps and libraries now support it.

This means we can target just netstandard2.0 (mostly). We also target net472 because we have special handling for RefEmit scenarios.

[LTS]: https://github.com/dotnet/core/blob/e2f22a7106860c0e5dc98bb36dc648a779944ad5/microsoft-support.md#net-core-releases


src/Microsoft.VisualStudio.Composition/Microsoft.VisualStudio.Composition.csproj
```diff
@@ -12,26 +12,18 @@
...
     <PackageReference Include="AsyncUsageAnalyzers" Version="1.0.0-alpha003" PrivateAssets="all" />
     <PackageReference Include="Microsoft.VisualStudio.Validation" Version="15.0.82" />
     <PackageReference Include="System.Composition" Version="1.0.31" />
     <PackageReference Include="System.Reflection.Emit" Version="4.3.0" />
     <PackageReference Include="System.Reflection.Metadata" Version="1.3.0" />
     <PackageReference Include="Roslyn.Diagnostics.Analyzers" Version="1.2.0-beta2" PrivateAssets="all" />
   </ItemGroup>
-  <ItemGroup Condition=" '$(TargetsDesktop)' == 'true' ">
+  <ItemGroup>
     <AdditionalFiles Include="PublicAPI.Shipped.txt" />
```
ikeras: Was a little surprised these files didn't seem to change in this PR despite the deletions and typeforwards, but if that's right, great!

AArnott: The type forwards were in a different assembly, so they wouldn't be picked up by this project's api txt files.

PR #151, opened 2019-08-06
Fix symbol archival for NetFx attributes

AArnott:


AArnott:
This will be available from nuget.org as 15.8.118

PR #173, opened 2020-01-30
BUG 1055784 : Adding traces to collect additional information while r…

ankitvarmait:
…esolving generic type

Nullable type arguments causing failures while creating a generic type, Throwing custom exception including all trace information collected to track which package causing this issue
Collect


src/Microsoft.VisualStudio.Composition/Diagnostic/DiagnosticInfoCollector.cs
```diff
@@ -0,0 +1,31 @@
+﻿// Copyright (c) Microsoft. All rights reserved.
+
+namespace Microsoft.VisualStudio.Composition.Diagnostic
+{
+    using System.Text;
+
+    public class DiagnosticInfoCollector
```
ZoeyR: This type should be internal. We don't want to modify public API for this.

src/Microsoft.VisualStudio.Composition/Reflection/TypeRef.cs
```diff
@@ -208,8 +220,18 @@ internal Type ResolvedType
...
+                        using (var genericTypeArguments = this.GetResolvedTypeArray(this.GenericTypeArguments, out bool isNullableTypeRefRequested))
                         {
+                            //Generic type does not support creation with nullable argument
+                            if (isNullableTypeRefRequested)
+                            {
+                                this.diagnosticInfoCollector.Collect($"Total GenericTypeArguments = {this.GenericTypeArguments.Length} for ResolvedType = {this.resolvedType}");
+                                this.diagnosticInfoCollector.Collect($"FullName = {this.FullName}");
+                                this.diagnosticInfoCollector.Collect($"AssemblyName = {this.AssemblyName}");
+
+                                throw new ArgumentException($"TypeArguments for the type parameters {genericTypeArguments.Value} of the current generic type should not be null. {this.diagnosticInfoCollector.GetAllInformation()}");
```
ZoeyR: Add a test verifying that we get this argument exception

src/Microsoft.VisualStudio.Composition/Diagnostic/DiagnosticInfoCollector.cs
```diff
@@ -0,0 +1,34 @@
...
+        {
+            return new DiagnosticInfoCollector();
+        }
+
+        public void Collect(string message)
+        {
+            this.messages.AppendLine(message);
+        }
+
+        public string GetAndClearInformation()
```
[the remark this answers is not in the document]
ankitvarmait: Thanks for feedback !
Updated code to collect information only when Type is null and that would not allow to collect information for all type resolver. Also Used List to avoid allocation issue. Please provide your feedback if can be pushed. .

src/Microsoft.VisualStudio.Composition/Diagnostic/DiagnosticInfoCollector.cs
```diff
@@ -0,0 +1,25 @@
...
+        private readonly List<string> messages = new List<string>();
+
+        internal void Collect(string message)
+        {
+            this.messages.Add(message);
+        }
+
+        internal string GetAndClearInformation()
+        {
+            var response = string.Join(Environment.NewLine, this.messages.ToArray());
```
AArnott: No reason to call `.ToArray()` on the List. `string.Join` has an overload that takes an `IEnumerable<T>`.

src/Microsoft.VisualStudio.Composition/Reflection/TypeRef.cs
```diff
@@ -388,17 +420,27 @@ public bool Equals(Type other)
...
             for (int i = 0; i < typeRefs.Length; i++)
             {
                 result.Value[i] = typeRefs[i].ResolvedType;
+
+                this.DiagnosticInfoCollector.Collect($"arg[{i}]={result.Value[i]}");
+                if (result.Value[i] == null && !isNullableTypeRefRequested)
+                {
+                    isNullableTypeRefRequested = true;
+                    this.DiagnosticInfoCollector.Collect($"arg[{i}] is expected to be not null.");
+                    this.DiagnosticInfoCollector.Collect($"arg[{i}] Details: {typeRefs[i].DiagnosticInfoCollector.GetAndClearInformation()}");
```
AArnott: This looks like it could snowball into a _very_ large string. If you take my earlier suggestion and avoid creating strings and instead store the data as flags and collections I think you can dramatically reduce the risk of significant GC pressure being added by the change.

ankitvarmait:
Sent different pull request

PR #174, opened 2020-02-06
Add nullable ref type annotations

AArnott:


src/Microsoft.VisualStudio.Composition.AppHost/CreateComposition.cs
```diff
@@ -171,7 +171,7 @@ public override bool Execute()
                 if (!string.IsNullOrEmpty(this.DgmlOutputPath))
                 {
                     configuration.CreateDgml().Save(this.DgmlOutputPath);
-                    this.writtenFiles.Add(this.DgmlOutputPath);
+                    this.writtenFiles.Add(this.DgmlOutputPath!);
```
maorleger, line 174: I didn't think you'd need ! here since you check string.IsNullOrEmpty on line 171

src/Microsoft.VisualStudio.Composition.AppHost/CreateComposition.cs
```diff
@@ -390,7 +390,7 @@ private Assembly CurrentDomain_AssemblyResolve(object sender, ResolveEventArgs a
         /// <param name="minimumVersion">Minimum version to consider.</param>
         /// <param name="assembly">loaded assembly.</param>
         /// <returns>true if assembly was loaded.</returns>
-        private bool Probe(string filePath, Version minimumVersion, out Assembly assembly)
+        private bool Probe(string filePath, Version minimumVersion, [NotNullWhen(true)] out Assembly? assembly)
```
maorleger, line 393: Neat! I have so much to learn...

src/Microsoft.VisualStudio.Composition/ByValueEquality+AssemblyNameComparer.cs
```diff
@@ -51,29 +51,26 @@ public bool Equals(AssemblyName x, AssemblyName y)
...
+                    && Equals(x.Version, y.Version)
                     && string.Equals(x.CultureName, y.CultureName)
                     && ByValueEquality.Buffer.Equals(x.GetPublicKeyToken(), y.GetPublicKeyToken());
             }
 
-            public int GetHashCode(AssemblyName obj)
-            {
-                return obj.Name.GetHashCode();
-            }
+            public int GetHashCode(AssemblyName obj) => obj.Name?.GetHashCode() ?? 0;
```
maorleger, line 73: Is AssemblyName.Name ever null? I wasn't sure from the docs

src/Microsoft.VisualStudio.Composition/ByValueEquality.cs
```diff
@@ -13,17 +13,19 @@ namespace Microsoft.VisualStudio.Composition
...
 
-        internal static IEqualityComparer<IReadOnlyDictionary<TKey, TValue>> Dictionary<TKey, TValue>(IEqualityComparer<TValue> valueComparer = null)
+        internal static IEqualityComparer<IReadOnlyDictionary<TKey, TValue>> Dictionary<TKey, TValue>(IEqualityComparer<TValue>? valueComparer = null)
+            where TKey : notnull
         {
             return DictionaryEqualityComparer<TKey, TValue>.Get(valueComparer);
         }
 
         internal static IEqualityComparer<IReadOnlyDictionary<TKey, ImmutableHashSet<TValue>>> DictionaryOfImmutableHashSet<TKey, TValue>()
+            where TKey : notnull
```
maorleger, line 28: I needed something just like this, and was unable to find a solution that didn't limit to reference types (but it clearly exists) - where is `notnull` coming from?

maorleger (commented):
🎉 this is awesome. Lots of changes so I'll need a few passes but asking some questions just for my own learning. Feel free to ignore

src/Microsoft.VisualStudio.Composition/ByValueEquality.cs
maorleger: Nevermind, I found it in the docs. But I'm not sure why it wasn't available for me in the VSLSP repo - maybe I just wasn't using the right constraint... anyway ignore me :smile:

src/Microsoft.VisualStudio.Composition.AppHost/CreateComposition.cs
AArnott: And you'd be right too, when targeting `netcoreapp3.1`. But we also target `netstandard2.0` and `net472` which apparently lack annotations on `string.IsNullOrEmpty`, so the C# compiler doesn't recognize that method's behavior for null value tracing. So I have to add this character to suppress a warning for the older frameworks.

src/Microsoft.VisualStudio.Composition/ByValueEquality+AssemblyNameComparer.cs
AArnott: I've never seen it `null` nor expect it to ever be. But [the annotations](https://source.dot.net/#System.Private.CoreLib/AssemblyName.cs,917bceeed23fe2ae) on `netcoreapp3.1` say it could be, so we have to check or use `!` to suppress the warning.

maorleger (approved):
This is a massive PR but I assume minimal functionality changes (compile time nullability annotations and runtime null checks where they make sense) so I skimmed most of it - if there's something specific you wanted me to look more closely at feel free to let me know, otherwise LGTM!

PR #175, opened 2020-02-07
Should throw exception when given type is not found and returns requested type in the request.

ankitvarmait:
Should throw exception when given type is not found and returns requested type in the request.
Unit test case

### DIFF
--- a/src/tests/Microsoft.VisualStudio.Composition.Tests/Reflection/TypeRefTests.cs
+++ b/src/tests/Microsoft.VisualStudio.Composition.Tests/Reflection/TypeRefTests.cs
@@ -48,7 +48,7 @@ namespace Microsoft.VisualStudio.Composition.Tests.Reflection
         }
 
         [Fact]
-        public void ThrowArgumentExceptionWhenArgIsNull()
+        public void ThrowTypeLoadExceptionWhenArgIsNull()
         {
             var testGuid = new Guid("00000000-0000-0000-0000-000000000001");
             var loadSystemAssembly = AppDomain.CurrentDomain.GetAssemblies().FirstOrDefault(assembly => assembly.FullName.Contains("mscorlib"));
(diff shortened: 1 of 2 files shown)


src/tests/Microsoft.VisualStudio.Composition.Tests/Reflection/TypeRefTests.cs
```diff
@@ -47,6 +47,28 @@ public void EqualsChecksMvidEquality()
...
+                CodeBase = loadSystemAssembly.CodeBase
+            };
+            var assemblyIdentity = new StrongAssemblyIdentity(loadedAssemblyName, testGuid);
+
+            var typeRefNullableArgument2 = TypeRef.Get(TestUtilities.Resolver, assemblyIdentity, 0x02000001, "Fake assembly", TypeRefFlags.None, 0, ImmutableArray<TypeRef>.Empty, false, ImmutableArray<TypeRef>.Empty, null);
+
+            var typeRef = TypeRef.Get(TestUtilities.Resolver, assemblyIdentity, 0x02000001, typeof(Dictionary<,>).FullName, TypeRefFlags.None, 0, new[] { typeRefNullableArgument2 }.ToImmutableArray(), false, new[] { typeRefNullableArgument2 }.ToImmutableArray(), null);
+
+            var actualException = Assert.Throws<TypeLoadException>(() => typeRef.Resolve());
+            Assert.NotNull(actualException);
```
AArnott: tip: you don't need to assert that `actualException` is not null. `Assert.Throws` guarantees to return a non-null result.

ankitvarmait: Thank you for providing feedback, I have updated the test case.

ankitvarmait:
g

PR #183, opened 2020-10-02
Add ExportProvider.ReleaseExport methods to avoid memory leak for non-shared parts

### LINKED ISSUE #135: Avoid memory leak for IDisposable/NonShared exports
Consider the sample below. When the exported type implements `IDisposable` the container keeps a reference to _all_ created values, even if they are _non-shared_. This easily causes significant memory leaks.

What is the recommended pattern here?

```c#
using System;
using System.ComponentModel.Composition;
using System.Reflection;
using System.Threading.Tasks;
using Microsoft.VisualStudio.Composition;
using CreationPolicy = System.ComponentModel.Composition.CreationPolicy;

namespace MefTest
{
    [Export(typeof(ViewModel)), PartCreationPolicy(CreationPolicy.NonShared)]
    public class ViewModel : IDisposable
    {
        private byte[] bytes;
        public ViewModel() => bytes = new byte[100000]; // Leak
        public void Dispose() {}
    }

    class Program
    {
        static void Main(string[] args)
        {
            var t = InitMef();
            t.Wait();

            var container = t.Result;

            for (var i = 0; i < 100; i++)
            {
                var model = container.GetExportedValue<ViewModel>();
                model.Dispose(); // Container keeps a reference here!
            }

            // NOTE: I want to keep my container.
            // How do I release the memory allocated above?
            // container.Dispose();
        }

        private static async Task<ExportProvider> InitMef()
        {
            var discovery = PartDiscovery.Combine(new AttributedPartDiscoveryV1(Resolver.DefaultInstance));

            var catalog = ComposableCatalog.Create(Resolver.DefaultInstance)
                .AddParts(await discovery.CreatePartsAsync(Assembly.GetExecutingAssembly()))
                .WithCompositionService();

            var config = CompositionConfiguration.Create(catalog);
            var epf = config.CreateExportProviderFactory();
            var exportProvider = epf.CreateExportProvider();

            return exportProvider;
        }
    }
}
```

AArnott:
Like ".NET MEF" (System.ComponentModel.Composition), VS-MEF did and still does hold strong references to any non-shared parts so they can be disposed with the container. This fix follows the API design of .NET MEF to mitigate memory leaks from doing so by adding ExportProvider.ReleaseExport(s)​ methods. When you get an Export​ or Lazy<T>​ where T is a non-shared part, passing this to the ReleaseExport method will:

1. Immediately dispose of the non-shared part and any more non-shared parts created for its import tree.
1. Remove the strong reference to this non-shared part from the container, allowing it to be GC'd once you release all references to it.

Note that this means you must call Release when you're done with the export -- not immediately after creating it. 

Note this does make some breaking API changes but only in the internal hosting APIs that we don't expect customers to be directly using, and we have repeatedly done this in the past with no complaints. To help make this clear and make such changes more freely in the future, I go ahead and _remove_ these and some related APIs from the public API surface area.

Fixes #135


maorleger (commented):
I'll review the rest first thing tomorrow

src/Directory.Build.props
```diff
@@ -28,7 +28,7 @@
   </PropertyGroup>
   <ItemGroup>
     <PackageReference Include="Nerdbank.GitVersioning" Version="3.2.31" PrivateAssets="all" />
-    <PackageReference Include="StyleCop.Analyzers" Version="1.1.118" PrivateAssets="all" />
+    <PackageReference Include="StyleCop.Analyzers" Version="1.2.0-beta.205" PrivateAssets="all" />
```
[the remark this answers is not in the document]
AArnott: Ya, it's OK. It doesn't impact what the customer sees as our own nuget dependencies, and StyleCop.Analyzers comes out with stable releases _very_ slowly. C# 8, which we use, requires a beta stylecop.analyzers in order to avoid false warnings.

src/Microsoft.VisualStudio.Composition/ExportProvider+DisposableLazy`2.cs
```diff
@@ -0,0 +1,26 @@
...
+    using System;
+
+    public partial class ExportProvider
+    {
+        private interface INonSharedLazy
+        {
+            NonSharedExport NonSharedExport { get; }
+        }
+
+        private class NonSharedLazy<T, TMetadata> : Lazy<T, TMetadata>, INonSharedLazy
```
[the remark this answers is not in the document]
AArnott: Fixed

src/Microsoft.VisualStudio.Composition/ExportProvider.cs
```diff
@@ -1097,12 +1231,33 @@ public object GetValueReadyToRetrieveExportingMembers()
             public void Dispose()
             {
                 this.isDisposed = true;
+
+                if (this.sharingBoundary is null && this.nonSharedPartOwner is null)
```
maorleger, line 1235: Should we use `this.IsNonShared` instead of `this.sharingBoundary is null` ?

maorleger (approved):
LGTM! A few questions but feel free to ignore

src/Microsoft.VisualStudio.Composition/ExportProvider.cs
```diff
@@ -374,6 +374,73 @@ where importDefinition.ExportConstraints.All(c => c.IsSatisfiedBy(export.Definit
...
+                disposable.Dispose();
+            }
+        }
+
+        /// <inheritdoc cref="ReleaseExport(Export)"/>
+        public void ReleaseExport<T>(Lazy<T> export)
+        {
+            Requires.NotNull(export, nameof(export));
+
+            if (export is INonSharedLazy nonSharedExport)
```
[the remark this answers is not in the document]
AArnott: The Lazy itself isn't disposed, as it is not IDisposable. But we extract from the Lazy the Export, which we then get the PartLifecycleTracker from and dispose it. That Tracker disposes of the value that was returned from the `Lazy<T>`. Calling .Value after this point will throw if it hadn't been run earlier because the value factory we gave it can't produce a value after the export was disposed.

src/Microsoft.VisualStudio.Composition/ExportProvider.cs
AArnott: Sure. It's equivalent but if you find it easier to read, that's cool.

src/Microsoft.VisualStudio.Composition/ExportProvider.cs
```diff
@@ -989,6 +1097,16 @@ protected internal abstract class PartLifecycleTracker : IDisposable
             /// </remarks>
             private HashSet<PartLifecycleTracker> deferredInitializationParts;
 
+            /// <summary>
+            /// A lazily-initialized collection of all transitive non-shared parts that should be disposed with this non-shared part.
+            /// </summary>
+            private HashSet<PartLifecycleTracker> nonSharedChildParts;
```
maorleger, line 1103: Thank you for walking me through the changes 

Assuming I understood correctly that a DAG of non-shared parts get flattened into a single level, is it possible (and I'm not sure it is) for me as a consumer to want to release an export that was maybe 3 levels deep and expect only its children to be released as well (but instead the entire group (siblings, ancestors, etc) gets released)? 

I _think_ the answer is no because I would likely be importing and holding onto the root and wouldn't release any transitive dependencies - just release the root when I'm finished. 

But.... Just wanted to check anyway.

AArnott: Good question, and correct deduction. The only `Export` or `Lazy<T>` exposed to the caller is the "root" non-shared part. So the only thing they could release is the root level one.

PR #197, opened 2021-02-09
Secure nuget packages via packages.lock.json

AArnott:


AArnott:
This was cherry-picked into v16.9 by a3f025f53121a1354d1c1a7

PR #199, opened 2021-02-10
Add install and uninstall scripts to analyzer package.

### LINKED ISSUE #198: Composition Analyzer package does not install when using packages.config
This analyzer package is missing the install.ps1 and uninstall.ps1 scripts that allow it to work in packages.config projects. It will only work for `<PackageReference>`.

The following error is logged in VS's NuGet Package Manager:

> Could not install package 'Microsoft.VisualStudio.Composition.Analyzers 16.9.6-alpha'. You are trying to install this package into a project that targets '.NETFramework,Version=v4.7.2', but the package does not contain any assembly references or content files that are compatible with that framework. For more information, contact the package author.

For more information on this, see [Analyzer Conventions - Install and Uninstall scripts](https://docs.microsoft.com/en-us/nuget/guides/analyzers-conventions#install-and-uninstall-scripts).

JoeRobich:
Fixes #198

### DIFF
--- a/src/Microsoft.VisualStudio.Composition.Analyzers/Microsoft.VisualStudio.Composition.Analyzers.csproj
+++ b/src/Microsoft.VisualStudio.Composition.Analyzers/Microsoft.VisualStudio.Composition.Analyzers.csproj
@@ -33,6 +33,10 @@
     </EmbeddedResource>
   </ItemGroup>
 
+  <ItemGroup>
+    <None Update="tools\*.ps1" CopyToOutputDirectory="Always" Pack="true" PackagePath="" />
+  </ItemGroup>
+
   <Target Name="PackBuildOutputs" DependsOnTargets="SatelliteDllsProjectOutputGroup;DebugSymbolsProjectOutputGroup">
     <ItemGroup>
       <TfmSpecificPackageFile Include="$(TargetPath)" PackagePath="analyzers\dotnet\" />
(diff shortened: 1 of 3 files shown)


src/Microsoft.VisualStudio.Composition.Analyzers/Microsoft.VisualStudio.Composition.Analyzers.csproj
```diff
@@ -33,6 +33,10 @@
     </EmbeddedResource>
   </ItemGroup>
 
+  <ItemGroup>
+    <None Update="tools\*.ps1" CopyToOutputDirectory="Always" Pack="true" PackagePath="" />
```
AArnott, line 37: We don't need to copy the scripts to the output directory at all. And to set it to `Always` breaks the VS up-to-date check. Can you remove that attribute?

Also, aren't these scripts supposed to go into the `tools` folder in the package? But I see you dropping them into the package root directory.

Here is a sample:
https://github.com/microsoft/vs-threading/blob/162a4b621f9ac4e8e1417d448f445dd11cf03e74/src/Microsoft.VisualStudio.Threading.Analyzers.CodeFixes/Microsoft.VisualStudio.Threading.Analyzers.CodeFixes.csproj#L22

Your use of `None` instead of `Content` should be fine though.

AArnott (changes requested):
Thanks for contributing.

src/Microsoft.VisualStudio.Composition.Analyzers/Microsoft.VisualStudio.Composition.Analyzers.csproj
JoeRobich: Updated. Thanks!

AArnott (approved):
Thanks

PR #201, opened 2021-04-21
Updated secure-development-tools.yml for compliance

milopezc:
Adding automated compliance tools to the build pipeline.


AArnott (approved):
Please queue a build directly on the devdiv pipeline before merging this to confirm this change is good. The file you modified here does not run in the PR build, so the fact that it passes is insufficient validation.

PR #212, opened 2021-05-19
Migrate MEF localization from XLF to LCL

milopezc:
Migrate MEF localization from XLF to LCL. See: https://devdiv.visualstudio.com/DevDiv/_workitems/edit/1274729


AArnott (approved):
Will you be reaching out to the loc team to advise them that we're ready for LCLs to be inserted?

PR #213, opened 2021-06-17
Reduce delegate allocations from method groups

stephentoub:
Using the VS allocation profiler to watch VS startup, it's allocating ~230K `Func<>` delegates, almost all of which are coming from Microsoft.VisualStudio.Composition.  After this commit, that number drops to ~15K.

### DIFF
--- a/src/Microsoft.VisualStudio.Composition/Configuration/CachedCatalog.cs
+++ b/src/Microsoft.VisualStudio.Composition/Configuration/CachedCatalog.cs
@@ -56,6 +56,10 @@ namespace Microsoft.VisualStudio.Composition
 
         private class SerializationContext : SerializationContextBase
         {
+            private Func<IImportSatisfiabilityConstraint?>? readIImportSatisfiabilityConstraint;
+            private Func<ImportDefinitionBinding>? readImportDefinitionBinding;
+            private Func<ExportDefinition>? readExportDefinition;
+
             internal SerializationContext(BinaryReader reader, Resolver resolver)
                 : base(reader, resolver)
             {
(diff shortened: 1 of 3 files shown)


src/Microsoft.VisualStudio.Composition/Configuration/CachedCatalog.cs
```diff
@@ -125,23 +129,29 @@ private void Write(ComposablePartDefinition partDefinition)
                 }
             }
 
+            private Func<ImportDefinitionBinding> ReadImportDefinitionBindingDelegate => this.readImportDefinitionBinding ??= this.ReadImportDefinitionBinding;
+
+            private Func<ExportDefinition> ReadExportDefinitionDelegate => this.readExportDefinition ??= this.ReadExportDefinition;
+
+            private Func<IImportSatisfiabilityConstraint?> ReadIImportSatisfiabilityConstraintDelegate => this.readIImportSatisfiabilityConstraint ??= this.ReadIImportSatisfiabilityConstraint;
+
```
AArnott, line 137: Was there some particular reason you lazily initialize these in properties rather than initialize the fields in the ctor and use them directly?

AArnott (approved):

src/Microsoft.VisualStudio.Composition/Configuration/CachedCatalog.cs
stephentoub: I'm happy to change it if you'd prefer.  I wasn't sure about whether there other use cases where all of these may not be used,  in which case the proactive allocation could be a deoptimization. Of that's not a concern,  I prefer the simplicity and efficiency of just allocating them all up front into readonly fields.

AArnott: I'm sure we'll end up initializing all of them, yes. Do you have to adjust this to initialize into `readonly` fields?

stephentoub: I'll push another commit.

AArnott (approved):
Thanks!

davkean:
@stephentoub Does this match what you reduced:

image

stephentoub:
> Does this match what you reduced

No.  At least in what I was profiling (opening VS, clicking the link to open without code, then closing VS), that particular allocation was low on the list.  I just reprofiled, and in that scenario it's representing only ~4K of the ~224K `Func<>` allocations:
image

This PR primarily addresses these, which represent ~210K of the ~224K.
image

I hadn't got to the ones lower on the list yet, but it seems like you're profiling something else that makes it pop?  It should be easy to remove the `Func<>` allocation here; it'll be harder to eliminate the associated allocation to capture the relevant state.  I can submit a PR to do the former.

stephentoub:
@davkean, can you share what scenario you're tracing?  For reference, after I apply this PR, comparatively this is what `Func<>` looks like for me when profiling startup:
image

AArnott:
I suspect `GetExportedValueHelper` shows up more in the MEF-heavy code paths. @stephentoub focused on the startup path (that's hit so long as MEF loads), whereas `GetExportedValueHelper` is hit proportional to the number of activated MEF parts.
That said, `GetExportedValueHelper` creates a `Func<object>` for each call, and this is explicitly by design. It is required in order to fit into `Lazy<T>`, which takes no args for the value factory and quite a few variables must be captured in the closure of that delegate. I don't think we'll find optimizations in that method, but I welcome second looks from others.

davkean:
This is [trace](https://microsoft.sharepoint.com/teams/VSIDEPerformance/Shared%20Documents/Forms/AllItems.aspx?id=%2Fteams%2FVSIDEPerformance%2FShared%20Documents%2FGeneral%2FTraces%2F17%2E0%2E0%20Preview%202%2E0%20%5B31422%2E448%2Emain%5D%2FSolution%5F006%5F0936%5FGeneral%5Fnocache&p=true&originalPath=aHR0cHM6Ly9taWNyb3NvZnQuc2hhcmVwb2ludC5jb20vOmY6L3QvVlNJREVQZXJmb3JtYW5jZS9Fa0MyekJXblE0RkZvYUJSdGp5WmRIVUJGTXhzNzU4bnRYRDd1NHhhZDBhaDRnP3J0aW1lPWtuck1wX0E1MlVn) from opening a 1000 project solution.

PR #245, opened 2021-07-26
Lazy initialize deferred initialization parts

davkean:
This allocates around 50 MB opening a large solution, this is an attempt at only allocating when actually needed for property imports:

image


AArnott (approved):
Looks good.

PR #258, opened 2021-09-08
Fix `PartDiscovery` assembly loading

AArnott:
- Fix `PartDiscovery` to use `IAssemblyLoader`
- Fix nullable ref annotation on `IAssemblyLoader.LoadAssembly(string, string?)`


AArnott:
CC: @sarda-devesh

milopezc (approved):

PR #271, opened 2021-11-25
Mefx like command line tool

sarda-devesh:


AArnott:
It looks like you somehow changed the chmod attributes of the `init.ps1` file:
image

Can you please revert that? You can probably do it with `git checkout origin/main -- :/init.ps1`

AArnott:
nit: can you rename the `doc/Images` directory to `doc/images`? We use lowercase directory names as a convention. 
On Windows, a case-only git rename can be tricky, but you should be able to do it like this:

```
git rm --cached doc/Images/*
git add doc/images/*
```

src/VS.Mefx/ConfigCreator.cs
```diff
@@ -0,0 +1,491 @@
...
+
+        /// <summary>
+        /// Initializes a new instance of the <see cref="ConfigCreator"/> class.
+        /// </summary>
+        /// <param name="options">The arguments inputted by the user.</param>
+        public ConfigCreator(CLIOptions options)
+        {
+            this.AssemblyPaths = new List<string>();
+            this.CachePaths = new List<string>();
+            this.PartInformation = new Dictionary<string, ComposablePartDefinition>();
```
AArnott, line 43: These first three assignments could be moved to property initializers, and the properties then made read only. For example:

```cs
internal List<string> AssemblyPaths { get; } = new();
```

src/VS.Mefx/ConfigCreator.cs
```diff
@@ -0,0 +1,491 @@
...
+
+        /// <summary>
+        /// Gets or sets the paths to the assembly files we want to read.
+        /// </summary>
+        private List<string> AssemblyPaths { get; set; }
+
+        /// <summary>
+        /// Gets or sets the paths to the cache files we want to read.
+        /// </summary>
+        private List<string> CachePaths { get; set; }
```
AArnott, line 122: tip: private properties often make me ask why not just use a field, which can be used in more contexts (e.g. `out` parameters).

src/VS.Mefx/ConfigCreator.cs
```diff
@@ -0,0 +1,491 @@
...
+
+        /// <summary>
+        /// Method to get the details about a part, i.e. the part Definition, given its name.
+        /// </summary>
+        /// <param name="partName"> The name of the part we want to get details about.</param>
+        /// <returns><see cref="ComposablePartDefinition"/> associated with the given part if it is
+        /// present in the catalog and null otherwise.</returns>
+        public ComposablePartDefinition? GetPart(string partName)
```
AArnott, line 130: Avoid using `public` modifier in a non-public type unless the member implements an interface. Otherwise in code reviews it leads to concern about adding public API when it isn't really public anyway.

src/VS.Mefx/ConfigCreator.cs
```diff
@@ -0,0 +1,491 @@
...
+        {
+            if (!this.PartInformation.ContainsKey(partName))
+            {
+                return null;
+            }
+
+            return this.PartInformation[partName];
```
AArnott, line 137: This results in two dictionary lookups. You can optimize this to just one by using this syntax:

```cs
this.PartInformation.TryGetValue(partName, out ComposablePartDefinition? partDefinition);
return partDefinition;
```

src/VS.Mefx/ConfigCreator.cs
```diff
@@ -0,0 +1,491 @@
...
+        /// <returns>A Task object when all the assembly have between loaded in and configured.</returns>
+        public async Task Initialize()
+        {
+            CustomAssemblyLoader customLoader = new CustomAssemblyLoader(this.Options.ErrorWriter);
+            Resolver customResolver = new(customLoader);
+            var nugetDiscover = new AttributedPartDiscovery(customResolver, isNonPublicSupported: true);
+            var netDiscover = new AttributedPartDiscoveryV1(customResolver);
+            PartDiscovery discovery = PartDiscovery.Combine(customResolver, netDiscover, nugetDiscover);
+
+            if (this.AssemblyPaths.Count() > 0)
```
AArnott, line 152: `Count()` is an extension method that will enumerate every element in the collection to count them. But `List<T>.Count` is a property that can readily return the answer, and would be the most performant option in this case.

When `Count` is *not* available as a property, use `Any()` instead of `Count() > 0` because it will only enumerate at most one element.

src/VS.Mefx/ConfigCreator.cs
```diff
@@ -0,0 +1,491 @@
...
+                        error.Message);
+                    this.Options.ErrorWriter.WriteLine(cacheReadError);
+                }
+            }
+        }
+
+        /// <summary>
+        /// Method to store the parts read from the input files into a cache for future use.
+        /// </summary>
+        private async Task SaveToCache()
```
AArnott, line 296: All methods returning awaitable types should use the `Async` suffix in its name.

src/VS.Mefx/ConfigCreator.cs
```diff
@@ -0,0 +1,491 @@
...
+            /// Loads the assembly with the assemblyName.
+            /// </summary>
+            /// <param name="assemblyName">The <see cref="AssemblyName"/> to read the assembly for.</param>
+            /// <returns>The loaded assembly with the given assemblyName.</returns>
+            public Assembly LoadAssembly(AssemblyName assemblyName)
+            {
+                if (DEBUG)
+                {
+#pragma warning disable CS0162 // Unreachable code detected
+                    this.Writer.WriteLine("Call to load assembly " + assemblyName.Name + " at " + DateTime.Now);
```
AArnott, line 416: We don't need this any more, do we? Same for your other logging in this class.

src/VS.Mefx/Commands/Command.cs
```diff
@@ -0,0 +1,77 @@
...
+            this.Options = arguments;
+        }
+
+        /// <summary>
+        /// Write the given lines in sorted order to the user.
+        /// </summary>
+        /// <param name="lines">The lines to output to the user.</param>
+        protected void WriteLines(List<string> lines)
+        {
+            lines.Sort();
```
AArnott, line 32: This has the side effect of reordering the collection the caller was holding, which per the doc comment on this method, isn't obvious.
Instead, you could copy the collection and sort that, or use the `OrderBy` extension method and then loop over that with `foreach`.

src/VS.Mefx/Commands/Command.cs
```diff
@@ -0,0 +1,77 @@
...
+            lines.ForEach(line => this.Options.Writer.WriteLine(line));
+        }
+
+        /// <summary>
+        /// Gets the composable catalog and configuration.
+        /// </summary>
+        protected ConfigCreator Creator { get; private set; } // Stores the catalog and config information for the input files.
```
AArnott, line 39: I think you can drop the `private set;` since only the ctor sets this property.
Also, why do you have the `//` code comment as well as the `<summary>` comment? Is the summary not correct or sufficient?

src/VS.Mefx/Commands/Command.cs
```diff
@@ -0,0 +1,77 @@
...
+        /// </summary>
+        protected CLIOptions Options { get; private set; } // The command line arguments specified by the user.
+
+        /// <summary>
+        /// Method to get the name of the given its definition.
+        /// </summary>
+        /// <param name="part"> The defintion of the part whose name we want.</param>
+        /// <param name="verboseLabel"> Label to add before the verbose description of the part.</param>
+        /// <returns> A string representing either the simple or verbose name of the part based
+        ///          on if verbose was specified as an input argument.</returns>
```
AArnott, line 52: Please don't indent the next line of `<returns>`. For multiline xml elements, consider putting the tags on their own lines and the text in the lines in between.

src/VS.Mefx/Commands/MatchChecker.cs
```diff
@@ -0,0 +1,446 @@
...
+        /// Method to perform matching on the input options and output the result to the user.
+        /// </summary>
+        public void PerformMatching()
+        {
+            if (this.Options.MatchParts == null || this.Options.MatchParts.Count == 0)
+            {
+                return;
+            }
+
+            if (this.Options.MatchParts.Count() == 2)
```
AArnott, line 38: Don't use the `Count()` extension method when a `Count` property exists.

src/VS.Mefx/Commands/MatchChecker.cs
```diff
@@ -0,0 +1,446 @@
...
+            {
+                string exportPartName = this.Options.MatchParts.ElementAt(0).Trim();
```
AArnott, line 40: Why use the `ElementAt` method when you can use the indexer (`this.Options.MatchParts[0].Trim()`)?

src/VS.Mefx/Commands/MatchChecker.cs
```diff
@@ -0,0 +1,446 @@
...
+                {
+                    total += 1;
+                    this.Options.Writer.WriteLine(result.Messages.First());
+                }
+                else
+                {
+                    for (int i = 0; i < result.Messages.Count; i++)
+                    {
+                        string failedConstraint = string.Format(
+                        CultureInfo.CurrentCulture,
```
AArnott, line 232: Missing indentation of arguments.

src/VS.Mefx/Commands/MatchChecker.cs
```diff
@@ -0,0 +1,446 @@
...
+
+            public string ImportingField { get; private set; }
+        }
+
+        private class PartExport
+        {
+            /// <summary>
+            /// Name of key to use when getting type of a given export.
+            /// </summary>
+            private static readonly string TypeKey = "ExportTypeIdentity";
```
AArnott, line 410: prefer `const` over `static readonly` for private members (and even some public ones).

src/VS.Mefx/Commands/PartInfo.cs
```diff
@@ -0,0 +1,251 @@
...
+        /// Method to output all the importing parts of a given contract name.
+        /// </summary>
+        /// <param name="contractName"> The contract name we want to analyze.</param>
+        private void ListTypeImporter(string contractName)
+        {
+            var importingParts = this.GetContractImporters(contractName);
+            string importPreview = string.Format(
+                        CultureInfo.CurrentCulture,
+                        Strings.ImportingContractsFormat,
+                        contractName);
```
AArnott, line 240: Please indent arguments only one level.

src/VS.Mefx/Commands/PartNode.cs
```diff
@@ -0,0 +1,145 @@
...
+            this.VerboseMessages = new HashSet<string> { message };
+            this.ImportRejects = new HashSet<PartEdge>();
+            this.RejectsCaused = new HashSet<PartEdge>();
+            this.Level = currLevel;
+            this.IsWhiteListed = false;
+
+            this.ExportingContracts = new List<string>();
+            foreach (var export in this.Part.ExportDefinitions)
+            {
+                if (export.Key == null)
```
AArnott, line 33: Dictionaries cannot have null keys. Why was this necessary?

src/VS.Mefx/Commands/PartNode.cs
```diff
@@ -0,0 +1,145 @@
...
+                }
+
+                this.ExportingContracts.Add(export.Value.ContractName);
+            }
+        }
+
+        /// <summary>
+        /// Gets the definition associated with the current part.
+        /// </summary>
+        public ComposablePartDefinition Part { get; private set; } // Represents the part associated with the current node
```
AArnott, line 45: Can we drop the `//` code comment since we already have a `/// <summary>` comment?

src/VS.Mefx/Commands/PartNode.cs
```diff
@@ -0,0 +1,145 @@
...
+        /// <summary>
+        /// Gets the verbose rejection messages associated with the current part.
+        /// </summary>
+        public HashSet<string> VerboseMessages { get; private set; } // Rejection Message(s) associated with the current part
+
+        /// <summary>
+        /// Gets the "children" of the current node, which represents parts that the current
+        /// node imports that have import issues themselves.
+        /// </summary>
+        public HashSet<PartEdge> ImportRejects { get; private set; }
```
AArnott, line 56: Use a property initializer here instead of initializing it in the constructor, here and where possible elsewhere.

src/VS.Mefx/Commands/PartNode.cs
```diff
@@ -0,0 +1,145 @@
...
+            return this.Part.Type.FullName!;
+        }
+
+        /// <summary>
+        /// Method to check if the given node imports any parts with import issues.
+        /// </summary>
+        /// <returns> A boolean indicating it the given node is a leaf node.</returns>
+        public bool IsLeafNode()
+        {
+            return this.ImportRejects.Count == 0;
```
AArnott, line 95: For these short one-liner methods, consider using the expression body syntax:

```cs
public bool IsLeafNode() => this.ImportsRejects.Count == 0;
```

And in fact this and some others may be better expressed as a property:

```cs
public bool IsLeafNode => this.ImportsRejects.Count == 0;
```

src/VS.Mefx/Commands/WhiteList.cs
```diff
@@ -0,0 +1,139 @@
...
+    using System;
+    using System.Collections.Generic;
+    using System.Globalization;
+    using System.IO;
+    using System.Text.RegularExpressions;
+
+    internal class WhiteList
+    {
+        // Constants associated with the Regex's for the expressions specified in the whitelist
+        private static readonly TimeSpan MaxRegexTime = new TimeSpan(0, 0, 5);
```
AArnott, line 15: `new TimeSpan(0, 0, 5)` leaves it difficult for the reader to know what units are used. I had to look it up.
Prefer syntax like `TimeSpan.FromSeconds(5)`.

AArnott (changes requested):
It looks very good. Thank you for putting this together. I have some suggestions, mostly around style.

src/VS.Mefx/VS.Mefx.csproj
```diff
@@ -0,0 +1,53 @@
+﻿<Project Sdk="Microsoft.NET.Sdk">
+
+  <PropertyGroup>
+    <OutputType>Exe</OutputType>
+    <TargetFrameworks>netcoreapp3.1;net5.0</TargetFrameworks>
```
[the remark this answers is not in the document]
sarda-devesh: > I don't think targeting two versions of .NET Core is necessary for this tool. Just `net5.0` should be fine. But we also want to target `net472` so we can analyze a MEF catalog that targets .NET Framework.

Hey Andrew, my PR was producing an error in the test pipeline when I was targeting just net5.0 and thus needed to add in netcoreapp3.1 to cause that error to disappear. Additionally, I tried to change the library to support the net472 framework but got an error saying that having a custom assembly load context is not supported (which is what we used to resolve the bug encountered during testing) in that framework.

AArnott:
Devesh, maybe this regressed when you removed the custom `IAssemblyLoader` in your tool, but when I try your tool at the CLI, it fails to load the assembly:

```
dotnet run -- --file C:\git\MEF\bin\Tests\Debug\net6.0\CarOne.dll C:\git\MEF\bin\Tests\Debug\net6.0\CarTwo.dll
```

It looks like .NET Core rejects it because only the assembly name is passed to `Assembly.Load`, and .NET Core ignores the codebase path. Did you have this working at some point in the past? If so, can you fix it?

AArnott:
Note: please `git pull` before making more changes as I've added to your PR.

sarda-devesh:
Hello, As you correctly guessed the issue was because of the fact that I removed the ```CustomAssemblyLoader```. This custom loader would create a custom load context and then load the requested files in the custom load context. The change meant that the program defaulted to using the ```Resolver.DefaultInstance``` which in turn used the ```StandardAssemblyLoader```. 

The issue is that the standard loader calls ```Assembly.Load``` which completes ignores the path of the file. To load the file from the file path, want to use ```Assembly.LoadFrom``` as that allows us to specify a file path. Thus I created another custom loader called ```LoadUsingPathFirst``` that loads the file using ```Assembly.LoadFrom``` if a file path is specified and using ```Assembly.Load``` if not. 

Doing so meant that when I run the following command on the command line ``` dotnet run --parts --file "<path to vs-mef folder>\bin\Tests\Debug\net6.0\CarOne.dll ``` I get the desired output of 

```
Parts in the Catalog are
CarOne.FirstCar
CarOne.InvalidType
CarOne.MoreMetadata
```

Since to load any assemblies in Mefx a file path must be specified, this means that we will always have a file path for every assembly we want to load.  This means that in reality, we will always load assemblies using the ```Assembly.LoadFrom``` method.  Going through the documentation, I found that 

> The load-from context contains assemblies for which the user provided a path that is not included in probing. LoadFrom, CreateInstanceFrom, and ExecuteAssembly are examples of methods that load by path.

I was wondering how much of an issue is the fact that LoadFrom doesn't use the default load context is in this case and what steps would you recommend to rectify it?

AArnott:
> I was wondering how much of an issue is the fact that LoadFrom doesn't use the default load context is in this case and what steps would you recommend to rectify it?

There are ways to rectify it (using `AppDomain.AssemblyResolve` event handlers) but it's very unnatural, and I don't see any problem with using the LoadFrom context in this case.

Note that some assemblies will actually load in the default context though -- assemblies that load implicitly because the ones we explicitly load reference them. Any assembly in the GAC or in the tool's own exe directory are candidates for loading automatically. That's as it should be, but just something to be aware of.

And speaking of assemblies loading implicitly, that's the next major work item to resolve in this new tool from my testing yesterday. I tried passing just one relatively simple dll from the VS installation to this tool and it failed with many assembly load errors. We need to be able to pass an .exe.config file to the tool so that it can resolve assemblies the same way a running app would. I have a (non-trivial) class that already does this for another purpose, so I'll wire it in as a switch to your tool and add to your PR.

AArnott:
@sarda-devesh In playing with this more, I find that *all* switches after `--file` are interpreted as file args to that one parameter instead of potentially switches of their own. In fact looking at your tests I see evidence that you may have encountered this as well, because `--file` is always the _last_ switch specified in your test data files. 
The `--file` switch is defined as a `List<string>?`, as is many other parameters. Does the CLI args parser library we're using even support specifying multiple parameters with more than one argument each?
I'm just information gathering at this point. We may have a few options moving forward.

AArnott (approved):
Thank you so much for this contribution.

PR #283, opened 2022-01-05
Make composition iterative instead of recursive

milopezc:
### Problem
Composition is proven to be a time consuming task in MEF, specially when errors are encountered during composition due to the recursive nature of the `CompositionConfiguration.Create` method which is called several times one for each error level found.

### Theory
Much of the processing done early in the method is expensive and could be done only once if taking an iterative approach. After the first run of validations the remaining salvaged parts can only be invalidated if they have a dependency with ExactlyOne cardinality on one of the invalidated parts. 

### Solution
The intention of this approach is to run that validation in an iterative way until we end up with a valid set of parts.

I added a Stopwatch prior and after this change to measure how long it took `CompositionConfiguration.Create` to complete and these were the results:

**Before change**

|Error Levels|Time 1|Time 2| Time 3|Average|
--- | --- | ---| ---| ---|
|1|1667|1545|1546|1586|
|3|2979|2953|3096|3009.3|
|6|5131|5157|5455|5247.6|
|9|7282|7075|7371|7242.6|

**After change**

|Error Levels|Time 1|Time 2| Time 3|Average|
--- | --- | ---| ---| ---|
|1|886|863|859|869.3|
|3|925|942|920|929|
|6|1049|1105|1138|1097.3|
|9|1107|1518|1158|1261|

Results show improvement of up to ~83% for 9 error levels and ~45% for the lower 1 level scenario.


src/Microsoft.VisualStudio.Composition/ComposedPart.cs
```diff
@@ -160,6 +160,57 @@ public IEnumerable<ComposedPartDiagnostic> Validate(IReadOnlyDictionary<Type, Ex
...
+                    }
+                    else
+                    {
+                        newSatisfyingExports.Add(export);
+                    }
+                }
+
+                if (invalidExportFound)
+                {
+                    this.SatisfyingExports = this.SatisfyingExports.Remove(pair.Key).Add(pair.Key, newSatisfyingExports);
```
AArnott, line 209: `Remove` followed by `Add` will create an intermediate `ImmutableDictionary` instance, resulting in several additional allocations. You can make this one operation by using `SetItem`.

src/Microsoft.VisualStudio.Composition/ComposedPart.cs
```diff
@@ -160,6 +160,57 @@ public IEnumerable<ComposedPartDiagnostic> Validate(IReadOnlyDictionary<Type, Ex
...
+        /// <returns>The set of <see cref="ComposedPartDiagnostic"/> with the encountered errors.</returns>
+        public IEnumerable<ComposedPartDiagnostic> CheckForInvalidatedParts(ImmutableHashSet<ComposablePartDefinition> invalidPartDefnitionsSet)
+        {
+            foreach (var pair in this.SatisfyingExports)
+            {
+                var importDefinition = pair.Key.ImportDefinition;
+
+                bool invalidExportFound = false;
+
+                List<ExportDefinitionBinding> newSatisfyingExports = new List<ExportDefinitionBinding>();
```
AArnott, line 177: Instead of allocating a new collection when most of the time we won't use it, consider leaving it `null` and crafting the code below to initialize it and populate it only with the elements to be removed. That should improve perf and GC pressure.
You could then remove your `invalidExportFound` local since you can tell there's something to change by testing `newSatisfyingExports` (probably renamed) against `null`.

src/Microsoft.VisualStudio.Composition/ComposedPart.cs
```diff
@@ -160,6 +160,57 @@ public IEnumerable<ComposedPartDiagnostic> Validate(IReadOnlyDictionary<Type, Ex
...
+                    // If the part is invalid it should be removed from the satisfying exports.
+                    if (invalidPartDefnitionsSet.Contains(export.PartDefinition))
+                    {
+                        // Signal that we found an export to remove so that the dictionary is updated
+                        invalidExportFound = true;
+
+                        switch (importDefinition.Cardinality)
+                        {
+                            // Only report error if the cardinality is exactly one
+                            // For mutiple or optional we just remove the satisfying export
```
AArnott, line 190: typo: "mutiple"

nit: inconsistent use of trailing periods in your code comments.

src/Microsoft.VisualStudio.Composition/ComposedPart.cs
```diff
@@ -160,6 +160,57 @@ public IEnumerable<ComposedPartDiagnostic> Validate(IReadOnlyDictionary<Type, Ex
...
+                            case ImportCardinality.ExactlyOne:
+                                yield return new ComposedPartDiagnostic(
```
AArnott, line 192: Using `yield return` in an `IEnumerable<T>` method guarantees more allocations, as the generator method now allocates a state machine and the caller has to allocate an `IEnumerator<T>`.
Since the caller doesn't need to *progressively* enumerate these errors (i.e. it's enough to get the whole collection at once), if you change the return type to a concrete collection type (e.g. `ComposedPartDiagnostic[]`) then you can allocate the error collection in this method only if errors are actually found and otherwise return `Array.Empty<ComposedPartDiagnostic>` (which returns a shared object). The concrete collection type means the caller will be compiled to call the public `GetEnumerator()` method that returns a struct rather than having to allocate an `IEnumerator<T>` object.

src/Microsoft.VisualStudio.Composition/ComposedPart.cs
```diff
@@ -160,6 +160,57 @@ public IEnumerable<ComposedPartDiagnostic> Validate(IReadOnlyDictionary<Type, Ex
...
+
+                foreach (var export in pair.Value)
+                {
+                    // If the part is invalid it should be removed from the satisfying exports.
+                    if (invalidPartDefnitionsSet.Contains(export.PartDefinition))
+                    {
+                        // Signal that we found an export to remove so that the dictionary is updated
+                        invalidExportFound = true;
+
+                        switch (importDefinition.Cardinality)
```
AArnott, line 187: nit: for a `switch` with only one `case` statement, would an `if` block be simpler?

src/Microsoft.VisualStudio.Composition/CompositionConfiguration.cs
```diff
@@ -150,27 +142,54 @@ public static CompositionConfiguration Create(ComposableCatalog catalog)
...
+                while (previousErrors.Count > 0)
                 {
-                    // If we can't identify the faulty parts but we still have errors, we have to just throw.
-                    throw new CompositionFailedException(Strings.FailStableComposition, ImmutableStack.Create<IReadOnlyCollection<ComposedPartDiagnostic>>(errors));
+                    compositionErrors = compositionErrors.Enqueue(previousErrors);
+
+                    // Get the salvaged parts
+                    var invalidParts = previousErrors.SelectMany(error => error.Parts);
+                    salvagedParts = salvagedParts.Except(invalidParts);
+                    var invalidPartDefnitionsSet = ImmutableHashSet.CreateRange(invalidParts.Select(p => p.Definition));
```
AArnott, line 163: typo: "Defnitions"

src/Microsoft.VisualStudio.Composition/CompositionConfiguration.cs
```diff
@@ -150,27 +142,54 @@ public static CompositionConfiguration Create(ComposableCatalog catalog)
...
+
+                // While we still find errors we validate the exports so that we remove all dependency failures
+                while (previousErrors.Count > 0)
                 {
-                    // If we can't identify the faulty parts but we still have errors, we have to just throw.
-                    throw new CompositionFailedException(Strings.FailStableComposition, ImmutableStack.Create<IReadOnlyCollection<ComposedPartDiagnostic>>(errors));
+                    compositionErrors = compositionErrors.Enqueue(previousErrors);
+
+                    // Get the salvaged parts
+                    var invalidParts = previousErrors.SelectMany(error => error.Parts);
```
AArnott, line 161: `invalidParts` is implicitly typed as `IEnumerable<T>`, and is enumerated *twice* below. That means you're executing the `SelectMany` twice. Also given that `IEnumerable<T>` is not a concrete collection type, each enumeration allocates an `IEnumerator<T>`. Can you add `.ToList()` to the end of this expression so that `SelectMany` executes only once and you end with a `List<T>` instead?

src/Microsoft.VisualStudio.Composition/CompositionConfiguration.cs
AArnott, line 163: Instead of creating an immutable collection here, consider using `List<T>` or simply `T[]` as that allocates just one or two objects instead of many as an immutable collection must.

src/Microsoft.VisualStudio.Composition/ComposedPart.cs
```diff
@@ -160,6 +160,57 @@ public IEnumerable<ComposedPartDiagnostic> Validate(IReadOnlyDictionary<Type, Ex
             }
         }
 
+        /// <summary>
+        /// Checks whether this part contains an import with <see cref="ImportCardinality.ExactlyOne"/> cardinality that has been invalidated,
+        /// which would invalidate this part as well.
+        /// </summary>
+        /// <param name="invalidPartDefnitionsSet">The set of definitions for the invalidated parts.</param>
+        /// <returns>The set of <see cref="ComposedPartDiagnostic"/> with the encountered errors.</returns>
+        public IEnumerable<ComposedPartDiagnostic> CheckForInvalidatedParts(ImmutableHashSet<ComposablePartDefinition> invalidPartDefnitionsSet)
```
AArnott, line 169: Taking `ImmutableHashSet` as the concrete type for a parameter ties the hands of your caller. Since the only use of this argument is to call `Contains`, accepting `ICollection<T>` should suffice. That will enable you to avoid creating an immutable collection in your caller, which has allocation benefits.

src/Microsoft.VisualStudio.Composition/CompositionConfiguration.cs
```diff
@@ -150,27 +142,54 @@ public static CompositionConfiguration Create(ComposableCatalog catalog)
...
-                var configuration = Create(salvagedCatalog);
-                return configuration.WithErrors(errors);
+                var finalCatalog = ComposableCatalog.Create(catalog.Resolver).AddParts(salvagedPartDefinitions);
+
+                var configuration = new CompositionConfiguration(
+                                finalCatalog,
+                                salvagedParts,
+                                metadataViewsAndProviders,
+                                compositionErrors,
+                                sharingBoundaryOverrides);
```
AArnott, line 182: Style: please only indent args by one indentation width.

src/Microsoft.VisualStudio.Composition/CompositionConfiguration.cs
```diff
@@ -150,27 +142,54 @@ public static CompositionConfiguration Create(ComposableCatalog catalog)
...
-                catalog,
-                parts,
-                metadataViewsAndProviders,
-                ImmutableStack<IReadOnlyCollection<ComposedPartDiagnostic>>.Empty,
-                sharingBoundaryOverrides);
+                                catalog,
+                                parts,
+                                metadataViewsAndProviders,
+                                ImmutableQueue<IReadOnlyCollection<ComposedPartDiagnostic>>.Empty,
+                                sharingBoundaryOverrides);
```
AArnott, line 192: Style: please only indent args by one indentation width.

AArnott (changes requested):
It's awesome to see how simple of a change this is but hear of the significant perf wins you get from it. Thanks for preparing this change!

src/Microsoft.VisualStudio.Composition/CompositionConfiguration.cs
milopezc: I went for using a HashSet since I want to use to see if it contains a given export I figured a HashSet would be faster on executing the Contains method.

src/Microsoft.VisualStudio.Composition/ComposedPart.cs
```diff
@@ -161,54 +174,58 @@ public IEnumerable<ComposedPartDiagnostic> Validate(IReadOnlyDictionary<Type, Ex
...
-                                    this,
-                                    Strings.RequiredImportHasBeenInvalidated,
-                                    GetDiagnosticLocation(pair.Key),
-                                    GetImportConstraints(pair.Key.ImportDefinition));
-
-                                break;
+                            if (errorsFound is null)
+                            {
+                                errorsFound = new List<ComposedPartDiagnostic>();
+                            }
```
AArnott, line 210: tip: this can be one line
```cs
errorsFound ??= new List<ComposedPartDiagnostic>();
```

src/Microsoft.VisualStudio.Composition/ComposedPart.cs
```diff
@@ -40,7 +42,18 @@ public ComposedPart(ComposablePartDefinition definition, ImmutableDictionary<Imp
...
+        {
+            get
+            {
+                return this.satisfyingExports;
+            }
+
+            private set
+            {
+                this.satisfyingExports = ImmutableDictionary.CreateRange(value);
+            }
```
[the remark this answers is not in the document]
AArnott: But you don't even need a setter, since we now have a field that you set directly. So this whole property can be written as:

```cs
public IReadOnlyDictionary<ImportDefinitionBinding, IReadOnlyList<ExportDefinitionBinding>> SatisfyingExports => this.satisfyingExports;
```

src/Microsoft.VisualStudio.Composition/ComposedPart.cs
```diff
@@ -160,6 +173,61 @@ public IEnumerable<ComposedPartDiagnostic> Validate(IReadOnlyDictionary<Type, Ex
             }
         }
 
+        /// <summary>
+        /// Find and removes any invalidated parts binded with an <see cref="ImportCardinality.ExactlyOne"/> from the <see cref="SatisfyingExports"/> dictionary.
```
AArnott, line 177: nit: "bound" instead of "binded"

AArnott: Also, describing the implementation details of what this part constitutes as an invalidation probably goes beyond what the caller cares about. It's probably enough for the summary to say "Removes satisfying exports coming from a given set of parts from this part's imports."
When it comes to that, the fact that the argument consists of _invalidated_ parts is not actually relevant to this method. So maybe call this RemoveSatisfyingExports and rename the parameter to `parts`. The doc comment for the parameter could then be changed to "The parts whose exports should not be imported by this part."

src/Microsoft.VisualStudio.Composition/ComposedPart.cs
```diff
@@ -160,6 +173,61 @@ public IEnumerable<ComposedPartDiagnostic> Validate(IReadOnlyDictionary<Type, Ex
...
+
+                foreach (var export in pair.Value)
+                {
+                    // If the part is invalid it should be removed from the satisfying exports.
+                    if (invalidPartDefinitionsSet.Contains(export.PartDefinition))
+                    {
+                        if (invalidExports is null)
+                        {
+                            invalidExports = new List<ExportDefinitionBinding>();
+                        }
```
AArnott, line 199: tip: Can be rewritten to just: `invalidExports ??= new();`

src/Microsoft.VisualStudio.Composition/CompositionConfiguration.cs
```diff
@@ -150,19 +142,49 @@ public static CompositionConfiguration Create(ComposableCatalog catalog)
...
+                var salvagedPartDefinitions = catalog.Parts;
+
+                List<ComposedPartDiagnostic> previousErrors = errors;
+                Stack<IReadOnlyCollection<ComposedPartDiagnostic>> stackedErrors = new Stack<IReadOnlyCollection<ComposedPartDiagnostic>>();
+
+                // While we still find errors we validate the exports so that we remove all dependency failures
+                while (previousErrors.Count > 0)
                 {
-                    // If we can't identify the faulty parts but we still have errors, we have to just throw.
-                    throw new CompositionFailedException(Strings.FailStableComposition, ImmutableStack.Create<IReadOnlyCollection<ComposedPartDiagnostic>>(errors));
```
AArnott, line 159: I don't see an equivalent place where you moved this code. Shouldn't we still notice and throw when a composition is invalid without identified parts?

AArnott (approved):
Looks good. Thanks!

PR #294, opened 2022-01-31
Fix deadlock in race condition

AArnott:
Ryan Molden analyzed a deadlock and the docs for `Monitor.Wait` and found that our call that includes a timeout means when the other lock holder takes more than 3 seconds to do its work and call `Monitor.PulseAll` that we'll still get `false` from `Monitor.Wait`. Then when we loop around to `Monitor.Wait` again, it will block for a `PulseAll` that may never recur.

The fix is to double-check the state before re-entering the Wait to see if we just missed the `PulseAll` call. If we did, we break out of the loop immediately.

I also add more to another code comment discussing thread concerns to make it clearer that it's doing what it is supposed to.


AArnott:
This fixes devdiv-1357805.

ryanmolden (approved):

milopezc (approved):

PR #297, opened 2022-02-03
Fix CI build failures

AArnott:
#271 broke our CI build at the sign verify task (which doesn't run in PR builds), and later in the policheck task. This PR fixes these issues.


doc/vsmefx.md
```diff
@@ -18,23 +18,25 @@ Use the `-?` parameter to get usage docs.
...
-  --whitelist <whitelist>          A file which lists the parts we expect to be rejected. [default: ]
-  --regex                          Treat the text in the whitelist file as regular expressions. [default: False]
-  --cache <cache>                  Specify the name of the output file to store the loaded parts. [default: ]
-  --match <match>                  Check relationship between given part which are provided in order: ExportPart ImportPart. [default: ]
-  --match-exports <match-exports>  List of fields in the export part that we want to consider. [default: ]
-  --match-imports <match-imports>  List of fields in the import part that we want to consider. [default: ]
-  --version                        Show version information
-  -?, -h, --help                   Show help and usage information
+  --verbose                                              An boolean option to toggle the detail level of the text output. [default: False]
+  --file <file>                                          Paths to assemblies (.dll or .exe) to include in the MEF catalog, or the path to a cache file written previously with the  switch. [default: ]
```
milopezc, line 22: nit: double whitespace before 'switch'

milopezc (approved):

doc/vsmefx.md
AArnott: Ah, that'll be a bug in the code too, since this was just copy/pasted from there. I'm expecting to revise the --file switch soon anyway, so we'll get it fixed then.

AArnott: Ah, that'll be a bug in the code too, since this was just copy/pasted from there. I'm expecting to revise the --file switch soon anyway, so we'll get it fixed then.

PR #314, opened 2022-03-30
Avoid func/capture allocations for non-lazy items 

davkean:
Non-lazy imports were paying for capture and Func<T> creation, despite the Func<T> being invoked immediately after it was created. Instead take a different path for the lazy versus non-lazy case so that only Lazy<T> exports pays for the allocations.

This was allocating about 100 MB opening and closing Orchard Core, but the end result will depend on how many lazy versus non-lazy items are in the composition graph:

image

This one was a bit tricky to approach to avoid leaking "this" and to handle export provider import and I took a few different paths, and ended up with this one. This one is much easier to read commit by commit.

### DIFF
--- a/src/Microsoft.VisualStudio.Composition/RuntimeExportProviderFactory+RuntimeExportProvider.cs
+++ b/src/Microsoft.VisualStudio.Composition/RuntimeExportProviderFactory+RuntimeExportProvider.cs
@@ -269,43 +269,50 @@ namespace Microsoft.VisualStudio.Composition
 
             private Func<object?> GetLazyExportedValue(RuntimeComposition.RuntimeImport import, RuntimeComposition.RuntimeExport export, RuntimePartLifecycleTracker? importingPartTracker)
             {
-                Requires.NotNull(import, nameof(import));
-                Requires.NotNull(export, nameof(export));
-
-                var exportingRuntimePart = this.composition.GetPart(export);
-
-                // Special case importing of ExportProvider
-                if (exportingRuntimePart.TypeRef.Equals(ExportProvider.ExportProviderPartDefinition.TypeRef))
-                {
-                    return () => this.NonDisposableWrapper.Value;
-                }
-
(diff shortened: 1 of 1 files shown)


src/Microsoft.VisualStudio.Composition/RuntimeExportProviderFactory+RuntimeExportProvider.cs
```diff
@@ -271,88 +267,124 @@ private object CreateExportFactory(RuntimePartLifecycleTracker importingPartTrac
...
+            }
+
+            /// <summary>
+            /// Called from <see cref="GetExportedValue(RuntimeComposition.RuntimeImport, RuntimeComposition.RuntimeExport, RuntimePartLifecycleTracker?, bool, out PartLifecycleTracker?)"/>
+            /// only, as an assisting method. See remarks.
+            /// </summary>
+            /// <remarks>
+            /// This method is separate from its one caller to avoid a csc.exe compiler bug
+            /// where it captures "this" in the closure for exportedValue, resulting in a memory leak
+            /// which caused one of our GC unit tests to fail.
```
AArnott, line 345: Is the compiler bug filed? Can we link to it from here?
Would a `static` local method suffice? Mostly just curious, but if it really only is meant to have the one caller, I tend to use local methods for things like that.

AArnott (approved):
Looks good to me. Yes, looking at it one commit at a time was a nice tip!

src/Microsoft.VisualStudio.Composition/RuntimeExportProviderFactory+RuntimeExportProvider.cs
davkean: This comment wasn't new, it was copied from prior code. I've inlined the method and added a comment referring to what I'm trying to avoid.

davkean:
@AArnott Can you merge this? I don't have approval.

milopezc (approved):

PR #326, opened 2022-06-07
Apply latest Library.Template

AArnott:
This creates an actual *merge* commit so that future Library.Template updates can be *much* easier. As a result, this PR looks huge by commit count, but that's a one-time hit and it is just bringing in history that has already mostly been in the repo by virtue of robocopy'ing updates from those commits in the past.


wade0016 (approved):

drewnoakes:
This appears to have accidentally closed some issues in this repo, where those issue IDs were from the Library.Template repo rather than this one.

AArnott:
Re-opened. Thank you, @drewnoakes.

PR #329, opened 2022-06-22
Handle 4-integer package versions in InsertConfigValues.ps1

matteo-prosperi:
This is sourced from https://github.com/microsoft/vs-streamjsonrpc/commit/679db2a7343675e138cb42469f227fe335668ed5


AArnott (approved):
It's a good change, although not currently necessary for this repo. It'll also come in with the next Library.Template update where I made the fix today.

PR #333, opened 2022-07-01
Generate consistent metadata view assembly names

davkean:
Investigating diffs in assembly loads and JIT time is difficult between builds of VS because the proxy assembly that contains the metadata views gets a new name every launch.

Consistently generate the names - so that the as long as the same assemblies are involved, the same name is generated.

This is an identical change I've made to ServiceBroker RPC proxies.

~(I've not actually tested this - as I cannot get MEF building at the moment on my box)~

### DIFF
--- a/src/Microsoft.VisualStudio.Composition/Configuration/MetadataViewGenerator.cs
+++ b/src/Microsoft.VisualStudio.Composition/Configuration/MetadataViewGenerator.cs
@@ -19,6 +19,8 @@ namespace Microsoft.VisualStudio.Composition
     using System.Reflection;
     using System.Reflection.Emit;
     using System.Security;
+    using System.Security.Cryptography;
+    using System.Text;
     using System.Threading;
     using Microsoft.VisualStudio.Composition.Reflection;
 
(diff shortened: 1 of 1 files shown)


src/Microsoft.VisualStudio.Composition/Configuration/MetadataViewGenerator.cs
```diff
@@ -94,12 +96,34 @@ internal static class MetadataViewGenerator
...
             return AssemblyBuilder.DefineDynamicAssembly(proxyAssemblyName, AssemblyBuilderAccess.Run);
         }
 
+        private static Guid GenerateGuidFromAssemblies(ImmutableHashSet<AssemblyName> assemblies)
+        {
+            // To make dll load and JIT comparisons between builds easier, we generate a
+            // consistent name for the proxy assembly based on the input assemblies involved.
+            using var algorithm = SHA256.Create();
+
+            foreach (AssemblyName name in assemblies.OrderBy(a => a.Name, StringComparer.Ordinal))
```
AArnott, line 111: Your guid generation must guarantee a unique guid for a given set of runtime assemblies.

I believe this should be `FullName`, because (on .NET Framework) there may be two Utilities assemblies loaded in the same appdomain if their fullname is unique.
Two AppDomains that load the same fullname shouldn't pose a problem because the guid only needs to be unique within one AppDomain.

On .NET, even FullName collisions are possible since one assembly may be loaded in multiple ALCs, but I guess that's a pre-existing bug in the code so no need to worry about that in your PR.

src/Microsoft.VisualStudio.Composition/Configuration/MetadataViewGenerator.cs
```diff
@@ -94,12 +96,34 @@ internal static class MetadataViewGenerator
...
+            {
+                var bytes = Encoding.UTF8.GetBytes(name.Name!);
```
AArnott, line 113: Why take the time to re-encode UTF16 bytes as UTF8? What if we just take the pointer to the UTF16 string we already have and encode the bytes in that memory directly? That would be faster.

AArnott (changes requested):
The overall change looks good. I've left a couple ideas for improvement.

src/Microsoft.VisualStudio.Composition/Configuration/MetadataViewGenerator.cs
davkean: Fixed.

davkean:
@AArnott Can I have another review on this?

AArnott (approved):

davkean:
@AArnott  thanks!

AArnott:
This should show up in VS day on Thursday (since it missed the cut-off for tonight's build and insertion).

PR #364, opened 2023-03-09
Version stabilization for 17. 6 release

ankitvarmait:
Version stabilization for 17. 6 release


AArnott (approved):
Looks good, provided you've tested with `dotnet pack -p:publicrelease=true` with this change applied. The PR build doesn't test this, but the CI build may break if issues that this `dotnet pack` command could also find are not fixed.

PR #420, opened 2023-09-08
Update to 17.8

RyanToth3:


AArnott (approved):

AArnott (changes requested):
Actually, why update this? It's already stable. For slow moving repos, we often don't bump this number.

RyanToth3:
I wasn't sure if we still wanted to update even though no changes happened. I'm fine abondoning the PR.

PR #433, opened 2023-11-30
Remove HasFlag usage as it's particularly slow on netfx

ToddGrun:
HasFlag is boxing the RuntimeImportFlags enum.

Image below is from a trace gathered when opening vs with a up to date MEF cache. IncPats is modified to just include Microsoft.VisualStudio.Composition, which indicates that this boxing is responsible for about 10% of allocations under callstacks including this namespace.

image


AArnott (approved):
First rate. Thanks.

PR #436, opened 2023-12-01
Partially revert earlier commit for always using ReadImmutableArray

ToddGrun:
It's better to let ReadList continue to do it's own thing and have ReadImmutableArray implemented separately.

One issue that I didn't notice in the profile after the earlier change was that ImmutableArray.Builder is a class, and for such small arrays, ended up allocating about as much as we were saving from not allocating duplicate arrays.

This changes the immutable array path to not use the builder the far majority of the time. In the case where the array length is > 4, we go ahead and use a builder, but do our best to reuse builders similar to metadataBuilder.

### DIFF
--- a/src/Microsoft.VisualStudio.Composition/Configuration/SerializationContextBase.cs
+++ b/src/Microsoft.VisualStudio.Composition/Configuration/SerializationContextBase.cs
@@ -47,6 +47,7 @@ namespace Microsoft.VisualStudio.Composition
         private readonly Func<TypeRef?> readTypeRefDelegate;
 
         private readonly ImmutableDictionary<string, object?>.Builder metadataBuilder = ImmutableDictionary.CreateBuilder<string, object?>();
+        private readonly Stack<ImmutableArray<TypeRef?>.Builder> typeRefBuilders = new Stack<ImmutableArray<TypeRef?>.Builder>();
 
         private readonly byte[] guidBuffer = new byte[128 / 8];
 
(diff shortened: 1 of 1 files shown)


src/Microsoft.VisualStudio.Composition/Configuration/SerializationContextBase.cs
```diff
@@ -677,16 +672,48 @@ protected ImmutableArray<T> ReadImmutableArray<T>(BinaryReader reader, Func<T> i
...
+                        return ImmutableArray.Create(itemReader(), itemReader(), itemReader(), itemReader());
+                    default:
+                        if (count > 0xffff)
+                        {
+                            // Probably either file corruption or a bug in serialization.
+                            // Let's not take untold amounts of memory by throwing out suspiciously large lengths.
+                            throw new NotSupportedException();
+                        }
+
+                        IEnumerable<T> items = Enumerable.Range(1, (int)count).Select(i => itemReader());
```
ToddGrun, line 714: I don't like this, let me try caching a list of builders instead.

ToddGrun: Cleaned it up to not use Enumerable.Range anymore (this had a couple issues outlined in the 2nd commit description).

src/Microsoft.VisualStudio.Composition/Configuration/SerializationContextBase.cs
```diff
@@ -677,16 +673,62 @@ protected ImmutableArray<T> ReadImmutableArray<T>(BinaryReader reader, Func<T> i
...
+                        return ImmutableArray.Create(itemReader(), itemReader(), itemReader());
+                    case 4:
+                        return ImmutableArray.Create(itemReader(), itemReader(), itemReader(), itemReader());
+                }
+
+                if (count > 0xffff)
+                {
+                    // Probably either file corruption or a bug in serialization.
+                    // Let's not take untold amounts of memory by throwing out suspiciously large lengths.
+                    throw new NotSupportedException();
```
AArnott, line 713: nit: NotSupportedException sounds a bit misleading. An InvalidOperationException with a message that contains the suspicious `count` value would be far more useful if this ever showed up in telemetry or logs.

AArnott (approved):
Looks good. Thanks.

src/Microsoft.VisualStudio.Composition/Configuration/SerializationContextBase.cs
AArnott: Can you update the PR description to match the final iteration?

src/Microsoft.VisualStudio.Composition/Configuration/SerializationContextBase.cs
ToddGrun: This was a direct copy from ReadList/ReadArray

AArnott: Fair enough. :)

PR #478, opened 2024-05-20
MessagePack serializer for vs-mef types

ankitvarmait:
MessagePack serializer for vs-mef types


ankitvarmait:
Will send a separate PR to remove old code

src/Microsoft.VisualStudio.Composition/Configuration/NetFxAdapters.cs
```diff
@@ -13,9 +13,12 @@ namespace Microsoft.VisualStudio.Composition
     using System.Reflection;
     using System.Text;
     using System.Threading.Tasks;
+    using MessagePack;
+    using Microsoft.VisualStudio.Composition.Formatter;
     using Microsoft.VisualStudio.Composition.Reflection;
     using MefV1 = System.ComponentModel.Composition;
 
+    [MessagePackObject]
```
AArnott, line 21: This is a `static` class. As such, it cannot be serialized because there is never an object for it.

src/Microsoft.VisualStudio.Composition/Configuration/Formatters/ImportSatisfiabilityConstraintFormatter.cs
```diff
@@ -0,0 +1,112 @@
...
+{
+    using System.Collections.Immutable;
+    using System.Globalization;
+    using MessagePack;
+    using MessagePack.Formatters;
+    using Microsoft.VisualStudio.Composition.Reflection;
+
+#pragma warning disable CS3001 // Argument type is not CLS-compliant
+
+    public class ImportSatisfiabilityConstraintFormatter : IMessagePackFormatter<IImportSatisfiabilityConstraint>
```
AArnott, line 13: What you've done here appears to be a manual version of [what `UnionAttribute` automates](https://github.com/MessagePack-CSharp/MessagePack-CSharp?tab=readme-ov-file#union). I suggest you apply `UnionAttribute` on `IImportSatisfiabilityConstraint`, and then focus on ensuring that all the derived types can be formatted (ideally via `[MessagePackObject]` attributes).

src/Microsoft.VisualStudio.Composition/Configuration/Formatters/ParameterRefFormatter.cs
```diff
@@ -0,0 +1,40 @@
...
+namespace Microsoft.VisualStudio.Composition.Formatter
+{
+    using System.Collections.Immutable;
+    using MessagePack;
+    using MessagePack.Formatters;
+    using Microsoft.VisualStudio.Composition.Reflection;
+
+#pragma warning disable CS3001 // Argument type is not CLS-compliant
+
+    public class ParameterRefFormatter : IMessagePackFormatter<ParameterRef?>
```
AArnott, line 13: This formatter doesn't appear to need to exist. You can use the `SerializationConstructorAttribute` on the `ParameterRef` constructor that takes the two values you serialize here, and use `[Key(0)]` style attributes on the two fields, and it should work great as a `[MessagePackObject]`. You know, like I see you did for `DiscoveredParts`.

src/Microsoft.VisualStudio.Composition/Reflection/MemberRef.cs
```diff
@@ -12,7 +12,10 @@ namespace Microsoft.VisualStudio.Composition.Reflection
     using System.Runtime.InteropServices;
     using System.Text;
     using System.Threading.Tasks;
+    using MessagePack;
+    using Microsoft.VisualStudio.Composition.Formatter;
 
+    [MessagePackFormatter(typeof(MemberRefFormatter<MemberRef>))]
```
AArnott, line 18: I suspect `UnionAttribute` is a better choice for this base class instead of doing it yourself with a custom formatter.

src/Microsoft.VisualStudio.Composition/DiscoveredParts.cs
```diff
@@ -9,11 +9,16 @@ namespace Microsoft.VisualStudio.Composition
     using System.Linq;
     using System.Text;
     using System.Threading.Tasks;
+    using MessagePack;
+    using Microsoft.VisualStudio.Composition.Formatter;
 
+    [MessagePackObject]
     public class DiscoveredParts
     {
+        [Key(0)]
```
AArnott, line 18: You shouldn't be serializing `static` members. Only serialize instance members since they are the only ones that vary by object.

src/Microsoft.VisualStudio.Composition/Configuration/Formatters/MessagePackSerializerContext.cs
```diff
@@ -0,0 +1,145 @@
...
+        }
+
+        public MessagePackSerializerContext(int estimatedObjectCount, IFormatterResolver resolver, Resolver compositionResolver)
+            : base(GetIFormatterResolver(resolver))
+        {
+            this.serializingObjectTable = new ConcurrentDictionary<object, uint>(SmartInterningEqualityComparer.Default);
+            this.CompositionResolver = compositionResolver;
+        }
+
+        private ConcurrentDictionary<object, uint>? serializingObjectTable;
```
AArnott, line 41: No reason to use `ConcurrentDictionary` because serialization is never multi-threaded. And each distinct serialization job must necessarily have its own instance of this class.

AArnott (changes requested):
Overall, it looks good. Several pattern-related suggestions to make however.

src/Microsoft.VisualStudio.Composition/Configuration/CachedCatalog.cs
```diff
@@ -12,46 +12,29 @@ namespace Microsoft.VisualStudio.Composition
...
+    using MessagePack;
+    using MessagePack.Resolvers;
     using Microsoft.VisualStudio.Composition.Reflection;
 
     public class CachedCatalog
     {
         protected static readonly Encoding TextEncoding = Encoding.UTF8;
 
-        public Task SaveAsync(ComposableCatalog catalog, Stream cacheStream, CancellationToken cancellationToken = default(CancellationToken))
+        public async Task SaveAsync(ComposableCatalog catalog, Stream cacheStream, CancellationToken cancellationToken = default(CancellationToken))
```
AArnott, line 23: This is a breaking change because it will no longer be able to read the cache created by a prior version. That could be an issue for VS itself. We should think carefully about what the migration plan is. We _may_ want to leave these methods alone. 
And we probably want to expose whatever few MessagePack-related types would enable the app to serialize the objects itself, as will be necessary to allow object de-duping when serializing several catalogs in a row.

src/Microsoft.VisualStudio.Composition/RuntimeComposition.cs
```diff
@@ -62,6 +65,7 @@ private RuntimeComposition(IEnumerable<RuntimePart> parts, IReadOnlyDictionary<T
                 e => (IReadOnlyCollection<RuntimeExport>)e.ToImmutableArray());
         }
 
+        [MessagePackFormatter(typeof(MessagePackCollectionFormatter<RuntimePart>))]
         public IReadOnlyCollection<RuntimePart> Parts
```
BertanAygun, line 69: Is this an optmization or does MessagePack not handle collections internally? ie would this not work if RuntimePart has the formatter attribute on it?

src/Microsoft.VisualStudio.Composition/RuntimeComposition.cs
```diff
@@ -279,10 +284,13 @@ public class RuntimePart : IEquatable<RuntimePart>
 
             public MethodBase? ImportingConstructorOrFactoryMethod => this.ImportingConstructorOrFactoryMethodRef?.MethodBase;
 
+            [MessagePackFormatter(typeof(MessagePackCollectionFormatter<RuntimeImport>))]
```
BertanAygun, line 287: Nit: Similar question to above but I didn't think IReadOnlyList required a custom formatter.

src/Microsoft.VisualStudio.Composition/Configuration/Formatters/RuntimePartFormatter.cs
```diff
@@ -0,0 +1,71 @@
...
+                      importingMembers,
+                      exports!,
+                      onImportsSatisfiedMethods,
+                      sharingBoundary);
+        }
+
+        /// <inheritdoc/>
+        public void Serialize(ref MessagePackWriter writer, RuntimePart value, MessagePackSerializerOptions options)
+        {
+            options.Resolver.GetFormatterWithVerify<TypeRef>().Serialize(ref writer, value.TypeRef, options);
```
BertanAygun, line 50: I am wondering if a custom formatter is really needed here since MessagePack should be able to serialize ImportingConstrcutorOrFactoryMethodRef as null. Otherwise it looks like you are just using existing formatters.

test/Microsoft.VisualStudio.Composition.Tests/XunitExtensions/Mef3DiscoveryTestCaseRunner.cs
```diff
@@ -197,10 +197,12 @@ private static async Task RoundtripCatalogSerializationAsync(ComposableCatalog c
             var ms = new MemoryStream();
             catalogSerialization.SaveAsync(catalog, ms).Wait();
             ms.Position = 0;
-            var deserializedCatalog = await catalogSerialization.LoadAsync(ms, TestUtilities.Resolver);
 
             var before = new StringWriter();
             catalog.ToString(before);
+
+            var deserializedCatalog = await catalogSerialization.LoadAsync(ms, TestUtilities.Resolver);
```
AArnott, line 204: Is there some significant reason behind this move? Did the old placement quit working, or are you testing something specifically by moving this?

src/Microsoft.VisualStudio.Composition/Configuration/Formatters/TypeRefObjectFormatter.cs
```diff
@@ -0,0 +1,69 @@
...
+            writer.Write(value.FullName);
+            options.Resolver.GetFormatterWithVerify<TypeRefFlags>().Serialize(ref writer, value.TypeFlags, options);
+            writer.Write(value.GenericTypeParameterCount);
+
+            IMessagePackFormatter<ImmutableArray<TypeRef>> typeRefFormatter = options.Resolver.GetFormatterWithVerify<ImmutableArray<TypeRef>>();
+            typeRefFormatter.Serialize(ref writer, value.GenericTypeArguments, options);
+
+            writer.Write(value.IsShallow);
+
+            if (!value.IsShallow)
```
AArnott, line 56: Another case of conditional serialization violating the constant array length passed to the base constructor.

src/Microsoft.VisualStudio.Composition/Configuration/Formatters/RuntimeImportFormatter.cs
```diff
@@ -0,0 +1,132 @@
...
+                if (value.ImportingSiteTypeWithoutCollectionRef != value.ImportingSiteTypeRef)
+                {
+                    throw new ArgumentException($"{nameof(value.ImportingSiteTypeWithoutCollectionRef)} and {nameof(value.ImportingSiteTypeRef)} must be equal when {nameof(value.Cardinality)} is not {nameof(ImportCardinality.ZeroOrMore)}.", nameof(value));
+                }
+            }
+
+            options.Resolver.GetFormatterWithVerify<IReadOnlyCollection<RuntimeExport>>().Serialize(ref writer, value.SatisfyingExports, options);
+
+            MetadataDictionaryFormatter.Instance.Serialize(ref writer, value.Metadata, options);
+            if (value.IsExportFactory)
```
AArnott, line 126: Another case of conditional serialization introducing variability into the array length.

src/Microsoft.VisualStudio.Composition/Configuration/Formatters/RuntimeImportFormatter.cs
```diff
@@ -0,0 +1,132 @@
...
+            }
+            else
+            {
+                importingMember = options.Resolver.GetFormatterWithVerify<MemberRef?>().Deserialize(ref reader, options);
+            }
+
+            IMessagePackFormatter<TypeRef> typeRefFormatter = options.Resolver.GetFormatterWithVerify<TypeRef>();
+            TypeRef importingSiteTypeRef = typeRefFormatter.Deserialize(ref reader, options);
+            TypeRef importingSiteTypeWithoutCollectionRef =
+    cardinality == ImportCardinality.ZeroOrMore ? typeRefFormatter.Deserialize(ref reader, options) : importingSiteTypeRef;
```
AArnott, line 55: nit: indentation

src/Microsoft.VisualStudio.Composition/Configuration/Formatters/RuntimeImportFormatter.cs
```diff
@@ -0,0 +1,132 @@
...
+    internal class RuntimeImportFormatter : BaseMessagePackFormatter<RuntimeImport>
+    {
+        public static readonly RuntimeImportFormatter Instance = new();
+
+        private RuntimeImportFormatter()
+            : base(expectedArrayElementCount: 7)
+        {
+        }
+
+        public enum RuntimeImportFlags : byte
```
AArnott, line 21: `private`

src/Microsoft.VisualStudio.Composition/Configuration/Formatters/RuntimeImportFormatter.cs
```diff
@@ -0,0 +1,132 @@
...
+    cardinality == ImportCardinality.ZeroOrMore ? typeRefFormatter.Deserialize(ref reader, options) : importingSiteTypeRef;
+
+            IReadOnlyList<RuntimeExport> satisfyingExports = options.Resolver.GetFormatterWithVerify<IReadOnlyList<RuntimeExport>>().Deserialize(ref reader, options);
+            IReadOnlyDictionary<string, object?> metadata = MetadataDictionaryFormatter.Instance.Deserialize(ref reader, options);
+            IReadOnlyCollection<string?> exportFactorySharingBoundaries = isExportFactory
+                ? options.Resolver.GetFormatterWithVerify<IReadOnlyCollection<string>>().Deserialize(ref reader, options)
+                : ImmutableList<string>.Empty;
+
+            return importingMember == null
+                        ? new RuntimeComposition.RuntimeImport(
```
AArnott, line 64: two levels too much indentation.

src/Microsoft.VisualStudio.Composition/Configuration/Formatters/MessagePackSerializerContext.cs
```diff
@@ -0,0 +1,161 @@
...
+            this.CompositionResolver = compositionResolver;
+        }
+
+        public Resolver CompositionResolver { get; }
+
+        public void Dispose()
+        {
+            GC.SuppressFinalize(this);
+        }
+
```
AArnott, line 30: This doesn't seem like a reasonable cause for being disposable in the first place.

src/Microsoft.VisualStudio.Composition/Configuration/Formatters/MessagePackSerializerContext.cs
```diff
@@ -0,0 +1,161 @@
...
+            private int serializingObjectCounter;
+
+            internal DedupingResolver(IFormatterResolver inner)
+            {
+                this.inner = inner;
+            }
+
+            public IMessagePackFormatter<T>? GetFormatter<T>()
+            {
+                if (!typeof(T).IsValueType)
```
AArnott, line 78: I'm a touch worried that this is too aggressive a check. There are so many reference objects being serialized, and most of them are unlikely to ever appear twice. Yet it adds cost and size to every object in order to skip it later, which may never happen.
Also, it'll malfunction if `Equals` is poorly overridden. 
I wonder if instead we should test for specific types. Perhaps the same types that use `TryPrepareDeserializeReusableObject` in the `SerializationContextBase` class.

src/Microsoft.VisualStudio.Composition/Configuration/Formatters/MetadataDictionaryFormatter.cs
```diff
@@ -0,0 +1,363 @@
...
+        public static readonly MetadataDictionaryFormatter Instance = new();
+
+        private MetadataDictionaryFormatter()
+        {
+        }
+
+        /// <inheritdoc/>
+        public void Serialize(ref MessagePackWriter writer, IReadOnlyDictionary<string, object?> value, MessagePackSerializerOptions options)
+        {
+            writer.Write(value.Count);
```
AArnott, line 26: This is your one allowed written struct for the formatter.
What I think you meant was `writer.WriteMapHeader(value.Count);`, which would then give you license to write `value.Count * 2` more values.

src/Microsoft.VisualStudio.Composition/Configuration/Formatters/MetadataDictionaryFormatter.cs
```diff
@@ -0,0 +1,363 @@
...
+                        throw new NotSupportedException(string.Format(CultureInfo.CurrentCulture, Strings.UnsupportedFormat, objectType));
+                }
+
+                return response;
+            }
+        }
+
+        private enum ObjectType : byte
+        {
+            Null,
```
AArnott, line 337: I know the original serialization code had this, but as msgpack has a native representation for nil, I suggest we just use that.

src/Microsoft.VisualStudio.Composition/Configuration/Formatters/MetadataDictionaryFormatter.cs
```diff
@@ -0,0 +1,363 @@
...
+
+            void SerializeObject(ref MessagePackWriter messagePackWriter, object? value)
+            {
+                if (value is null)
+                {
+                    messagePackWriter.Write((byte)MetadataDictionaryFormatter.ObjectType.Null);
+                    return;
+                }
+
+                switch (value.GetType())
```
AArnott, line 52: If you drop `.GetType()`, then your null check cas be just a `case null:` statement.
And all the rest of your case statements get simpler too. See [this example](https://sharplab.io/#v2:CYLg1APgAgTAjAWAFBQMwAJboMLoN7LpGYAs6A6gE4CWALgKYAUA9gEYBW9AxrQPzoA3AIYAbAK70AlPkLE5AZwDudLgAt0jYeKkykc/cS5D59dADsxIkSFkG7mOAE5GAIgtWXkgNy37+1pT0QgDWPnp+REYm6ACClJRCAJ7oQvFJNuERclBOrqkJiZ5hWf6BIcVZUaaszMwi6DV1AGqiEhkl2bmNIi3a3r4lAUGhA8QAvr4TSGNAA==).

src/Microsoft.VisualStudio.Composition/Configuration/Formatters/MetadataDictionaryFormatter.cs
```diff
@@ -0,0 +1,363 @@
...
+
+                    case Type objectType when objectType == typeof(bool):
+                        ObjectType objectValueType = (bool)value ? ObjectType.BoolTrue : ObjectType.BoolFalse;
+                        messagePackWriter.Write((byte)objectValueType);
+                        break;
+
+                    case Type objectType when objectType == typeof(string):
+                        messagePackWriter.Write((byte)ObjectType.String);
+                        messagePackWriter.Write((string)value);
+                        break;
```
AArnott, line 77: You're writing twice for a single value, which breaks msgpack parsers. I suggest we fix this with a two pronged approach:
1. Don't use the `ObjectType` enum for any type for which msgpack has a matching primitive, because such primitive types can be detected while reading by virtue of the `MessagePackReader`, making the enum redundant for all those cases. Take care though when you write though that you use the explicit type named methods (e.g. `WriteUInt32`) instead of just `Write(uint)`, or else msgpack's type annotation will be 'lossy', which we can't afford. In the reader, you can build a switch statement around `reader.NextMessagePackType`.
2. For the types for which msgpack doesn't have a matching primitive, you'll need to call `WriteArrayHeader(2)` followed by the `ObjectType` enum value, and finally the metadatum value itself.

src/Microsoft.VisualStudio.Composition/Configuration/Formatters/MetadataDictionaryFormatter.cs
```diff
@@ -0,0 +1,363 @@
...
+                        var typeArraySubstitutionValue = (LazyMetadataWrapper.TypeArraySubstitution)value;
+                        messagePackWriter.Write((byte)ObjectType.TypeArraySubstitution);
+                        options.Resolver.GetFormatterWithVerify<IReadOnlyCollection<TypeRef?>>().Serialize(ref messagePackWriter, typeArraySubstitutionValue.TypeRefArray, options);
+
+                        break;
+
+                    default:
+                        var typeLessData = MessagePackSerializer.Typeless.Serialize(value);
+                        messagePackWriter.Write((byte)ObjectType.TypeLess);
+                        options.Resolver.GetFormatterWithVerify<byte[]>().Serialize(ref messagePackWriter, typeLessData, options);
```
AArnott, line 178: A formatter should _not_ call back into the toplevel `MessagePackSerializer.Serialize` method. Instead, it should call directly into a formatter. So something like this:

```cs
messagePackWriter.WriteArrayHeader(2);
messagePackWriter.Write((byte)ObjectType.Typeless);
TypelessFormatter.Instance.Serialize(ref writer, value, options);
```

AArnott (changes requested):
Much better. Several more improvements can be made. I look forward to the next iteration!

AArnott (changes requested):

src/Microsoft.VisualStudio.Composition/Configuration/Formatters/MetadataDictionaryFormatter.cs
```diff
@@ -12,16 +12,22 @@ namespace Microsoft.VisualStudio.Composition.Formatter;
 using Microsoft.VisualStudio.Composition.Reflection;
 using static Microsoft.VisualStudio.Composition.LazyMetadataWrapper;
 
+#pragma warning disable RS0016 // Add public types and members to the declared API
```
AArnott: We should not be hiding public API from the analyzer. Either we add it to the PublicAPI file or (in this case) we should make this internal. If making it internal breaks something, let me know and we'll find a way to fix it.

AArnott (changes requested):
I like all your other recent changes. I still see a bunch of modified files with the `namespace` block to statement change that is causing merge conflicts. I'll wait for that to be corrected and do one more pass over your PR.

src/Microsoft.VisualStudio.Composition/PartDiscoveryException.cs
```diff
@@ -5,11 +5,13 @@ namespace Microsoft.VisualStudio.Composition
 {
     using System;
     using System.Runtime.Serialization;
+    using MessagePack;
 
     /// <summary>
     /// An exception that may be thrown during MEF part discovery.
     /// </summary>
     [Serializable]
+    [MessagePackObject(true)]
```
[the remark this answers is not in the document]
ankitvarmait: I ended up creating the custom formatter. Please review it and let me know your feedback.

AArnott (approved):

AArnott (approved):
I recommend squashing this PR, since it has 99 commits in it.

PR #506, opened 2024-07-22
Reduce allocations

davkean:
This cleans up some minor allocations while reading the cache.


gundermanc (approved):

AArnott:
This is all good, but some is obsolete in light of #478 which will merge very shortly. Let's trim this PR down to avoid the code that is about to be removed.

AArnott (approved):

PR #507, opened 2024-07-29
Optimize the custom analyzer and deduping resolver

ankitvarmait:


src/Microsoft.VisualStudio.Composition/Configuration/Formatters/MessagePackSerializerContext.cs
```diff
@@ -104,30 +104,25 @@ internal MyDedupingResolver(IFormatterResolver inner)
...
-                    MessagePackReader provisionaryReader = reader.CreatePeekReader();
-                    ExtensionHeader extensionHeader = provisionaryReader.ReadExtensionFormatHeader();
-                    if (extensionHeader.TypeCode == ReferenceExtensionTypeCode)
-                    {
-                        int id = provisionaryReader.ReadInt32();
-                        reader = provisionaryReader;
-
-                        return (T)(this.owner.deserializedObjects[id] ?? throw new MessagePackSerializationException("Unexpected null element in shared object array. Dependency cycle?"));
-                    }
+                    reader.ReadExtensionFormatHeader();
```
AArnott, line 124: Your change here discards the extension typecode and reads 4 bytes, assuming it matches. This will malfunction when the value to be deserialized is a different extension than the one used by this particular formatter (which is allowed).

src/Microsoft.VisualStudio.Composition/Configuration/Formatters/MetadataObjectFormatter.cs
```diff
@@ -151,45 +153,80 @@ public void Serialize(ref MessagePackWriter messagePackWriter, object? value, Me
         }
     }
 
-    public object? Deserialize(ref MessagePackReader reader, MessagePackSerializerOptions options)
+    public ImmutableDictionary<string, object?> DeserializeMetadataObjects(ref MessagePackReader reader, MessagePackSerializerOptions options, int count)
```
AArnott, line 156: 💅🏼 Avoid using `public` on members that are not intended to be public API. 
In this case, you're in an `internal` type so ultimately this API is internal, making `public` here misleading. During a code review, public API impact merits more careful review than internal changes, so seeing this sets me up to immediately ask "Why is this *public*?" only to later realize "oh, it's not _really_ public, because it's in an `internal` type", so it's a time waste.

The only reason you should ever have `public` members on an `internal` type is when the C# language requires it, which is true for operator overloads and interface implementing members.

AArnott: I don't see this method used in this class at all. It serves only the class from which you removed the code that serves as this method body. How is this valuable? It increases the code churn in the code review, so there should be some value to it or IMO it should be reverted.

src/Microsoft.VisualStudio.Composition/Configuration/Formatters/MetadataObjectFormatter.cs
```diff
@@ -151,45 +153,80 @@ public void Serialize(ref MessagePackWriter messagePackWriter, object? value, Me
...
-        MessagePackType messagePackType = reader.NextMessagePackType;
+    private object? Deserialize(
+        ref MessagePackReader reader,
+        MessagePackSerializerOptions options,
+        IMessagePackFormatter<string> stringFormatter,
+        IMessagePackFormatter<TypeRef?> typeRefFormatter,
+        IMessagePackFormatter<Guid> guidFormatter,
+        IMessagePackFormatter<IReadOnlyList<TypeRef>> typeRefFormatterCollection)
+    {
+        return DeserializeObject(ref reader);
```
AArnott, line 193: 📖 I don't follow how this refactoring would improve perf. Let's sync on this.

src/Microsoft.VisualStudio.Composition/Configuration/Formatters/MetadataObjectFormatter.cs
```diff
@@ -151,45 +153,80 @@ public void Serialize(ref MessagePackWriter messagePackWriter, object? value, Me
...
+
+            for (int i = 0; i < count; i++)
+            {
+                dictionary.Add(stringFormatter.Deserialize(ref reader, options), this.Deserialize(ref reader, options, stringFormatter, typeRefFormatter, guidFormatter, typeRefFormatterCollection));
+            }
+        }
+        finally
         {
-            return null;
+            reader.Depth--;
```
AArnott, line 174: 👮🏼 Oops. You're decrementing where you didn't increment. This line is missing above the `try` block:

```cs
options.Security.DepthStep(ref messagePackReader);
```

Apparently the code you moved this from had this bug too.

src/Microsoft.VisualStudio.Composition/Reflection/TypeRef.cs
```diff
@@ -51,6 +51,11 @@ public class TypeRef : IEquatable<TypeRef>, IEquatable<Type>
         /// </summary>
         private ImmutableArray<TypeRef> baseTypes;
 
+        private static readonly Lazy<string> UnresolvableMetadataTokenString = new(() => Strings.UnresolvableMetadataToken);
```
AArnott, line 54: This adds overhead to acquiring the string, and you inevitably acquire the string anyway. It *does* store the retrieved string for reuse later, but it does so without concern for the culture set on the given thread, so you may return the wrong language.
There's another way to optimize the perf here so that you only retrieve the string in failure paths. I'll push the change.

AArnott (changes requested):

src/Microsoft.VisualStudio.Composition/Configuration/Formatters/MetadataObjectFormatter.cs
ankitvarmait: Only in case of loop

ankitvarmait: check at allother places

AArnott:
Closing because the baseline has been reverted.

PR #510, opened 2024-07-31
Revert "MessagePack serializer for vs-mef types"

ankitvarmait:
Reverts microsoft/vs-mef#478


ankitvarmait:
This PR reverts the use of MessagePack in VS-MEF as it increased load time for the existing MEF cache in Visual Studio scenarios. We will maintain a topic branch with these and upcoming changes to potentially re-enable it later.

PR #575, opened 2025-03-17
Include TPNs for 2025 Compliance

ankitvarmait:


AArnott (changes requested):
We already have a NOTICE file in each built package. It's generated by the build and appears to be identical to this file. Why would we also need to check it into the source tree?

ankitvarmait:
> We already have a NOTICE file in each built package. It's generated by the build and appears to be identical to this file. Why would we also need to check it into the source tree?

Thanks for the information! I double-checked the guidance, and it states that either package/root is sufficient, so we don’t need to include it here since it’s already in the NuGet package. I also verified the file in the package, and it matches. Appreciate your review! I’ve marked this PR as addressed

PR #581, opened 2025-05-07
Modify MetadataViewClassProvider.CreateProxy to potentially avoid allocation

ToddGrun:
Reuse the input metadata if it is already an ImmutableDictionary, avoiding allocations and cpu costs

This should go through a speedometer run to see if this improves the costs I'm currently seeing, or if there aren't many callers that are passing in an ImmutableDictionary.

image

### DIFF
--- a/src/Microsoft.VisualStudio.Composition/MetadataViewClassProvider.cs
+++ b/src/Microsoft.VisualStudio.Composition/MetadataViewClassProvider.cs
@@ -37,9 +37,9 @@ namespace Microsoft.VisualStudio.Composition
             ConstructorInfo? ctor = FindConstructor(metadataViewType.GetTypeInfo());
             Requires.Argument(ctor is not null, nameof(metadataViewType), "No public constructor with the required signature found.");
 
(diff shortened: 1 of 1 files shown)


ToddGrun:
@dotnet-policy-service rerun

ToddGrun:
Speedometer results came back good:

*** OLD ***
image
image

*** NEW ***
image
image

AArnott (approved):

ekoppel (approved):

weltkante:
Just out of interest, is this improving startup performance or is it actually a hot path at runtime? For the latter its worth getting rid of reflection overhead ('FindConstructor' and 'Invoke' in above performance charts) by creating a delegate that calls the constructor. Pays the reflection cost once and then lets the JIT take over.

AArnott:
@weltkante This will be mostly startup performance, but the function could theoretically be called many times during startup. And theoretically it could be with the same metadata view type, where the additional optimization you're talking about can be meaningful. 

I'll merge this as-is since it's a good improvement, but if @ToddGrun wants to go further I'll be happy to merge another PR.

PR #603, opened 2025-07-28
Add several new analyzers

### LINKED ISSUE #257: Analyzer could detect duplicate imports
I found a MEF part today that imported the same object via two constructor parameters, but with different names.

The same is possible for parameter injection.

While harmless, it's a little messy and may lead to confusion. An analyzer could detect this.

### LINKED ISSUE #287: Add analyzer to detect when a class has more than one `ImportingConstructor` attribute
A class I was modifying was long and the existing constructor wasn't at the top of the file, so I added another. I assume the error was written to the ComponentModelCache/Microsoft.VisualStudio.Default.err file, but neither compiling nor running the experimental instance gave me any indication about this error. I was just lucky that CI tests hit a code path where another component tried to import this class.

Anyway, if the analyzer checked for classes that have more than one `[ImportingConstructor]`, this type of error would be caught more quickly.

### LINKED ISSUE #317: Analyzer to ensure correct C# nullable annotation when AllowDefault = true
```c#
#nullable enable

[Import(AllowDefault = true)]
public IBar Bar { get; private set; }

[ImportingConstructor]
public Foo([Import(AllowDefault = true)] IBar bar)
{}
```

Both of these imports lead to potential runtime problems if a null value is not guarded against.

When in a `nullable` context, use of `AllowDefault = true` on a non-nullable type should be flagged with a diagnostic. A codefix could also be added to change the type to be nullable.

drewnoakes:
Fixes #257
Fixes #287
Fixes #317

New analyzers for:

- Missing `[ImportingConstructor]`
- Duplicate `[ImportingConstructor]`
- Duplicate `[Import]`
- Mismatched C# nullability and `AllowDefault`

Mostly coded using GitHub Copilot.

### DIFF
--- a/src/Microsoft.VisualStudio.Composition.Analyzers.CodeFixes/VSMEF004ExportWithoutImportingConstructorCodeFixProvider.cs
+++ b/src/Microsoft.VisualStudio.Composition.Analyzers.CodeFixes/VSMEF004ExportWithoutImportingConstructorCodeFixProvider.cs
@@ -1,0 +1,228 @@
+// Copyright (c) Microsoft Corporation. All rights reserved.
+// Licensed under the MIT license. See LICENSE file in the project root for full license information.
+
+namespace Microsoft.VisualStudio.Composition.Analyzers.CodeFixes;
+
+using System.Collections.Immutable;
+using System.Composition;
+using System.Linq;
+using System.Threading;
+using System.Threading.Tasks;
+using Microsoft.CodeAnalysis;
+using Microsoft.CodeAnalysis.CodeActions;
+using Microsoft.CodeAnalysis.CodeFixes;
+using Microsoft.CodeAnalysis.CSharp;
+using Microsoft.CodeAnalysis.CSharp.Syntax;
+
+/// <summary>
+/// Provides code fixes for VSMEF004: Exported type missing importing constructor.
+/// </summary>
+[ExportCodeFixProvider(LanguageNames.CSharp, Name = nameof(VSMEF004ExportWithoutImportingConstructorCodeFixProvider))]
+[Shared]
+public class VSMEF004ExportWithoutImportingConstructorCodeFixProvider : CodeFixProvider
+{
+    /// <inheritdoc/>
+    public sealed override ImmutableArray<string> FixableDiagnosticIds =>
+        ImmutableArray.Create(VSMEF004ExportWithoutImportingConstructorAnalyzer.Id);
+
+    /// <inheritdoc/>
+    public sealed override FixAllProvider GetFixAllProvider() => WellKnownFixAllProviders.BatchFixer;
+
+    /// <inheritdoc/>
+    public sealed override async Task RegisterCodeFixesAsync(CodeFixContext context)
+    {
+        var root = await context.Document.GetSyntaxRootAsync(context.CancellationToken).ConfigureAwait(false);
+        if (root is null)
+        {
+            return;
+        }
+
+        var diagnostic = context.Diagnostics.FirstOrDefault(d => d.Id == VSMEF004ExportWithoutImportingConstructorAnalyzer.Id);
+        if (diagnostic is null)
+        {
+            return;
+        }
+
+        var diagnosticSpan = diagnostic.Location.SourceSpan;
+        var constructorNode = root.FindNode(diagnosticSpan);
+
+        // Find the constructor declaration
+        var constructorDeclaration = constructorNode.FirstAncestorOrSelf<ConstructorDeclarationSyntax>();
+        if (constructorDeclaration is null)
+        {
+            return;
+        }
+
+        var classDeclaration = constructorDeclaration.FirstAncestorOrSelf<ClassDeclarationSyntax>();
+        if (classDeclaration is null)
+        {
+            return;
+        }
+
+        var semanticModel = await context.Document.GetSemanticModelAsync(context.CancellationToken).ConfigureAwait(false);
+        if (semanticModel is null)
+        {
+            return;
+        }
+
+        // Determine which MEF version to use based on existing attributes
+        var mefVersion = DetermineMefVersion(classDeclaration, semanticModel);
+
+        // Primary fix: Add [ImportingConstructor] attribute
+        var addAttributeAction = CodeAction.Create(
+            title: "Add [ImportingConstructor] attribute",
+            createChangedDocument: c => AddImportingConstructorAttributeAsync(context.Document, constructorDeclaration, mefVersion, c),
+            equivalenceKey: "AddImportingConstructorAttribute");
+
+        context.RegisterCodeFix(addAttributeAction, diagnostic);
+
+        // Secondary fix: Add parameterless constructor
+        var addConstructorAction = CodeAction.Create(
+            title: "Add parameterless constructor",
+            createChangedDocument: c => AddParameterlessConstructorAsync(context.Document, classDeclaration, c),
+            equivalenceKey: "AddParameterlessConstructor");
+
+        context.RegisterCodeFix(addConstructorAction, diagnostic);
+    }
+
+    private static async Task<Document> AddImportingConstructorAttributeAsync(
+        Document document,
+        ConstructorDeclarationSyntax constructorDeclaration,
+        MefVersion mefVersion,
+        CancellationToken cancellationToken)
+    {
+        var root = await document.GetSyntaxRootAsync(cancellationToken).ConfigureAwait(false);
+        if (root is null)
+        {
+            return document;
+        }
+
+        // Create the ImportingConstructor attribute
+        var attributeName = mefVersion == MefVersion.V1
+            ? "System.ComponentModel.Composition.ImportingConstructor"
+            : "System.Composition.ImportingConstructor";
+
+        var attribute = SyntaxFactory.Attribute(SyntaxFactory.ParseName(attributeName));
+        var attributeList = SyntaxFactory.AttributeList(SyntaxFactory.SingletonSeparatedList(attribute));
+
+        // Add the attribute to the constructor
+        var newConstructor = constructorDeclaration.WithAttributeLists(
+            constructorDeclaration.AttributeLists.Add(attributeList));
+
+        var newRoot = root.ReplaceNode(constructorDeclaration, newConstructor);
+        return document.WithSyntaxRoot(newRoot);
+    }
+
+    private static async Task<Document> AddParameterlessConstructorAsync(
+        Document document,
+        ClassDeclarationSyntax classDeclaration,
+        CancellationToken cancellationToken)
+    {
+        var root = await document.GetSyntaxRootAsync(cancellationToken).ConfigureAwait(false);
+        if (root is null)
+        {
+            return document;
+        }
+
+        // Create a parameterless constructor
+        var newConstructor = SyntaxFactory.ConstructorDeclaration(classDeclaration.Identifier)
+            .WithModifiers(SyntaxFactory.TokenList(SyntaxFactory.Token(SyntaxKind.PublicKeyword)))
+            .WithParameterList(SyntaxFactory.ParameterList())
+            .WithBody(SyntaxFactory.Block());
+
+        // Find the position to insert the constructor (after existing constructors)
+        var existingConstructors = classDeclaration.Members.OfType<ConstructorDeclarationSyntax>().ToList();
+        int insertIndex = 0;
+
+        if (existingConstructors.Any())
+        {
+            // Insert after the last constructor
+            var lastConstructor = existingConstructors.Last();
+            insertIndex = classDeclaration.Members.IndexOf(lastConstructor) + 1;
+        }
+        else
+        {
+            // Insert at the beginning of the class
+            insertIndex = 0;
+        }
+
+        var newMembers = classDeclaration.Members.Insert(insertIndex, newConstructor);
+        var newClass = classDeclaration.WithMembers(newMembers);
+
+        var newRoot = root.ReplaceNode(classDeclaration, newClass);
+        return document.WithSyntaxRoot(newRoot);
+    }
+
+    private static MefVersion DetermineMefVersion(ClassDeclarationSyntax classDeclaration, SemanticModel semanticModel)
+    {
+        // Check the class and its members for MEF attributes to determine which version to use
+        var classSymbol = semanticModel.GetDeclaredSymbol(classDeclaration);
+        if (classSymbol is null)
+        {
+            return MefVersion.V2; // Default to V2
+        }
+
+        // Check class-level attributes
+        foreach (var attribute in classSymbol.GetAttributes())
+        {
+            if (IsMefV1Attribute(attribute.AttributeClass))
+            {
+                return MefVersion.V1;
+            }
+
+            if (IsMefV2Attribute(attribute.AttributeClass))
+            {
+                return MefVersion.V2;
+            }
+        }
+
+        // Check member-level attributes
+        foreach (var member in classSymbol.GetMembers())
+        {
+            foreach (var attribute in member.GetAttributes())
+            {
+                if (IsMefV1Attribute(attribute.AttributeClass))
+                {
+                    return MefVersion.V1;
+                }
+
+                if (IsMefV2Attribute(attribute.AttributeClass))
+                {
+                    return MefVersion.V2;
+                }
+            }
+        }
+
+        return MefVersion.V2; // Default to V2 if no MEF attributes found
+    }
+
+    private static bool IsMefV1Attribute(INamedTypeSymbol? attributeType)
+    {
+        if (attributeType is null)
+        {
+            return false;
+        }
+
+        var namespaceName = attributeType.ContainingNamespace?.ToDisplayString();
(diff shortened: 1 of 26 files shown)


src/Microsoft.VisualStudio.Composition.Analyzers/README.md
```diff
@@ -5,3 +5,8 @@ Analyzers for MEF consumers to help identify common errors in MEF parts.
 Analyzer ID | Description
 --|--
 VSMEF001 | Ensures that importing properties define a `set` accessor.
+VSMEF002 | Detects mixing of MEF v1 and MEF v2 attributes on the same type.
+VSMEF004 | Ensures exported types have a parameterless constructor or importing constructor.
```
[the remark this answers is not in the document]
drewnoakes: That PR merged. Conflicts resolved.

test/Microsoft.VisualStudio.Composition.Analyzers.Tests/VSMEF006ImportNullabilityAnalyzerTests.cs
```diff
@@ -0,0 +1,421 @@
...
+    public async Task NullableImportWithoutAllowDefault_Warning()
+    {
+        string test = """
+            #nullable enable
+            using System.ComponentModel.Composition;
+
+            class Foo
+            {
+                [Import]
+                public string? {|#0:Value|} { get; set; }
```
AArnott, line 69: The trouble with this is removing the nullable annotation produces a C# warning because the (implicit) constructor doesn't guarantee the property is initialized with a non-null value.
This _can_ be addressed using either of these syntaxes:

```cs
[Import]
public required string Value { get; set; }
```

or 

```cs
[Import]
public string Value { get; set; } = null!;
```

I prefer the first option, though it requires a polyfill for .NET Framework/Standard folks.
The issue with this diagnostic would be that people won't know how to fix it, unless the diagnostic message volunteers the available workarounds.

src/Microsoft.VisualStudio.Composition.Analyzers/AnalyzerReleases.Unshipped.md
```diff
@@ -5,3 +5,7 @@
 Rule ID | Category | Severity | Notes
 --------|----------|----------|-------
 VSMEF003 | Usage | Warning | Exported type not implemented by exporting class
+VSMEF004 | Usage | Error | VSMEF004ExportWithoutImportingConstructor
+VSMEF005 | Usage | Error | VSMEF005MultipleImportingConstructors
+VSMEF006 | Usage | Warning | VSMEF006ImportNullability
+VSMEF007 | Usage | Warning | VSMEF007DuplicateImportAnalyzer
```
AArnott, line 11: Can we replace the values for the Notes column with something like what you have in index.md?
Also, I like your choice of severities here.

test/Microsoft.VisualStudio.Composition.Analyzers.Tests/VSMEF006ImportNullabilityAnalyzerTests.cs
```diff
@@ -0,0 +1,421 @@
+// Copyright (c) Microsoft Corporation. All rights reserved.
+// Licensed under the MIT license. See LICENSE file in the project root for full license information.
+
+using Microsoft.VisualStudio.Composition.Analyzers;
+using VerifyCS = CSharpCodeFixVerifier<Microsoft.VisualStudio.Composition.Analyzers.VSMEF006ImportNullabilityAnalyzer, Microsoft.VisualStudio.Composition.Analyzers.CodeFixes.VSMEF006ImportNullabilityCodeFixProvider>;
+
+public class VSMEF006ImportNullabilityAnalyzerTests
```
AArnott, line 7: All of these tests are based on importing reference types. We should test importing of value types and assert that none of them are ever expected to bear the `?` type suffix regardless of `AllowDefault` values.

src/Microsoft.VisualStudio.Composition.Analyzers/VSMEF005MultipleImportingConstructorsAnalyzer.cs
```diff
@@ -0,0 +1,103 @@
...
+            {
+                context.RegisterSymbolAction(context => AnalyzeSymbol(context), SymbolKind.NamedType);
+            }
+        });
+
+        static void AnalyzeSymbol(SymbolAnalysisContext context)
+        {
+            var symbol = (INamedTypeSymbol)context.Symbol;
+
+            // Only analyze classes
```
AArnott, line 63: why not structs? They can have too many importing constructors, can't they?

src/Microsoft.VisualStudio.Composition.Analyzers/VSMEF007DuplicateImportAnalyzer.cs
```diff
@@ -0,0 +1,167 @@
...
+                // Check if this constructor has [ImportingConstructor] or if it's the only constructor
+                bool isImportingConstructor = Utils.HasImportingConstructorAttribute(method) ||
+                    (namedType.Constructors.Length == 1 && Utils.HasInstanceExports(namedType));
+
+                if (isImportingConstructor)
+                {
+                    foreach (var parameter in method.Parameters)
+                    {
+                        var importAttribute = Utils.GetImportAttribute(parameter.GetAttributes());
+                        if (importAttribute is not null)
```
AArnott, line 102: An importing constructor parameter is still an import even if it lacks this attribute. Is this a blindspot in the analyzer?

src/Microsoft.VisualStudio.Composition.Analyzers/VSMEF007DuplicateImportAnalyzer.cs
```diff
@@ -0,0 +1,167 @@
...
+                context.ReportDiagnostic(Diagnostic.Create(
+                    Descriptor,
+                    import.Location,
+                    namedType.Name,
+                    import.Contract));
+            }
+        }
+    }
+
+    private static string? GetImportContract(AttributeData importAttribute, ITypeSymbol importType)
```
AArnott, line 131: It's actually far more complicated than this (see `ContractNameServices`), but since analyzers can't reference the library and I wouldn't want to reproduce all that logic here, a rough approximation is fine since it's not about what the contract name is _exactly_ as much as just identifying duplicates. But for clarity, that's probably worth commenting here.

src/Microsoft.VisualStudio.Composition.Analyzers.CodeFixes/VSMEF004ExportWithoutImportingConstructorCodeFixProvider.cs
```diff
@@ -0,0 +1,228 @@
...
+using Microsoft.CodeAnalysis;
+using Microsoft.CodeAnalysis.CodeActions;
+using Microsoft.CodeAnalysis.CodeFixes;
+using Microsoft.CodeAnalysis.CSharp;
+using Microsoft.CodeAnalysis.CSharp.Syntax;
+
+/// <summary>
+/// Provides code fixes for VSMEF004: Exported type missing importing constructor.
+/// </summary>
+[ExportCodeFixProvider(LanguageNames.CSharp, Name = nameof(VSMEF004ExportWithoutImportingConstructorCodeFixProvider))]
```
AArnott, line 20: What does the `Name` property on this attribute do? I've never bothered setting it.

src/Microsoft.VisualStudio.Composition.Analyzers.CodeFixes/VSMEF004ExportWithoutImportingConstructorCodeFixProvider.cs
```diff
@@ -0,0 +1,228 @@
...
+        }
+
+        var newMembers = classDeclaration.Members.Insert(insertIndex, newConstructor);
+        var newClass = classDeclaration.WithMembers(newMembers);
+
+        var newRoot = root.ReplaceNode(classDeclaration, newClass);
+        return document.WithSyntaxRoot(newRoot);
+    }
+
+    private static MefVersion DetermineMefVersion(ClassDeclarationSyntax classDeclaration, SemanticModel semanticModel)
```
AArnott, line 156: Remember that a MEF part may have no attributes from either of these namespaces at all, if the only MEF attributes it has are custom ones derived from an ExportAttribute.

test/Microsoft.VisualStudio.Composition.Analyzers.Tests/VSMEF004ExportWithoutImportingConstructorAnalyzerTests.cs
```diff
@@ -0,0 +1,657 @@
...
+                [System.Composition.ImportingConstructor]
+                public Service(string config) { }
+            }
+            """;
+
+        var expected = VerifyCS.Diagnostic().WithLocation(6, 12).WithArguments("Service");
+        await VerifyCS.VerifyCodeFixAsync(testCode, expected, fixedCode);
+    }
+
+    // TODO: Add test for second code action (parameterless constructor) when framework supports testing multiple code actions
```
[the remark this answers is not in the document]
drewnoakes: Thanks. Added non-skipped tests.

test/Microsoft.VisualStudio.Composition.Analyzers.Tests/VSMEF006ImportNullabilityAnalyzerTests.cs
```diff
@@ -0,0 +1,421 @@
...
+    [Fact]
+    public async Task NullablePropertyImportWithoutAllowDefault_Warning()
+    {
+        string test = """
+            #nullable enable
+            using System.ComponentModel.Composition;
+
+            class Foo
+            {
+                [Import]
```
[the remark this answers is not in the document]
drewnoakes: Oh, so a non-default value can still be null? Wow. I still lean towards adding this diagnostic as that seems like a pretty exotic case, and I think this adds value for the main case.

test/Microsoft.VisualStudio.Composition.Analyzers.Tests/VSMEF007DuplicateImportAnalyzerTests.cs
```diff
@@ -0,0 +1,409 @@
...
+            using System.ComponentModel.Composition;
+
+            [Export]
+            class Foo
+            {
+                [Import]
+                public string {|#0:Value1|} { get; set; }
+
+                [Import]
+                public string {|#1:Value2|} { get; set; }
```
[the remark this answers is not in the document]
drewnoakes: Is there an implicit order to those imports in such cases?

src/Microsoft.VisualStudio.Composition.Analyzers/VSMEF007DuplicateImportAnalyzer.cs
drewnoakes: Yes, fixed.

src/Microsoft.VisualStudio.Composition.Analyzers/VSMEF005MultipleImportingConstructorsAnalyzer.cs
drewnoakes: They can, I just haven't seen any use of value types with MEF so didn't think it worth adding these. Will do so now.

test/Microsoft.VisualStudio.Composition.Analyzers.Tests/VSMEF006ImportNullabilityAnalyzerTests.cs
drewnoakes: Thanks! Fixed. `int?` and `int` are now correctly treated as separate types, and the presence or absence of `AllowDefault` has no impact on the analyzer.

src/Microsoft.VisualStudio.Composition.Analyzers/VSMEF005MultipleImportingConstructorsAnalyzer.cs
drewnoakes: Actually, `ExportAttribute` is only allowed on classes and properties. Are you sure we want to expand support for value types?

src/Microsoft.VisualStudio.Composition.Analyzers.CodeFixes/VSMEF004ExportWithoutImportingConstructorCodeFixProvider.cs
drewnoakes: Best I can tell it's used in the UI, but then these values aren't that helpful. Removed.

src/Microsoft.VisualStudio.Composition.Analyzers/VSMEF005MultipleImportingConstructorsAnalyzer.cs
AArnott: Nope. That's fine.

test/Microsoft.VisualStudio.Composition.Analyzers.Tests/VSMEF006ImportNullabilityAnalyzerTests.cs
drewnoakes: It's easier to append ` = null!` for these, so I'll go that route for now. Understanding when `required` is available is complex, especially for in-proc VS components that are running on framework. I am not sure we could easily bake in knowledge of all the polyfills that people use. If the goal is to have the codefix produce immediately-compilable code, then `= null!` is safest.

src/Microsoft.VisualStudio.Composition.Analyzers/Utils.cs
```diff
@@ -1,26 +1,278 @@
...
-        /// <returns>The URL for the analyzer's documentation.</returns>
-        internal static string GetHelpLink(string analyzerId)
+        if (attributeType is null)
         {
-            return $"https://microsoft.github.io/vs-mef/analyzers/{analyzerId}.html";
+            return false;
         }
+
+        // Check if it's a direct Export attribute or derived from one
+        return IsAttributeOfType(attributeType, "ExportAttribute", MefV1AttributeNamespace.AsSpan()) ||
```
AArnott: What about [`InheritedExportAttribute`](https://catalog.apireview.net/catalog/a1f8acf311f6dc566ec9d8ea45b8b969)?

src/Microsoft.VisualStudio.Composition.Analyzers/Utils.cs
```diff
@@ -1,26 +1,278 @@
...
+        // Check instance members (properties, methods, fields) for Export attributes
+        foreach (ISymbol member in symbol.GetMembers())
+        {
+            // Skip static members and nested types
+            if (member.IsStatic || member is ITypeSymbol)
+            {
+                continue;
+            }
+
+            if (Utils.HasExportAttribute(member))
```
AArnott: Will this pick up on `InheritedExportAttribute` inherited from base members? I'm not sure whether an interface implementing method would inherit attributes from the interface member, or if this only applies to `override` members.

src/Microsoft.VisualStudio.Composition.Analyzers/Utils.cs
```diff
@@ -1,26 +1,278 @@
...
+    /// <param name="symbol">The type symbol to analyze.</param>
+    /// <returns><see langword="true"/> if the type has instance exports; otherwise, <see langword="false"/>.</returns>
+    /// <remarks>
+    /// This method checks for export attributes on the type itself or on its instance members (properties, methods, fields).
+    /// Static members with export attributes are ignored since they don't require type instantiation.
+    /// </remarks>
+    internal static bool HasInstanceExports(INamedTypeSymbol symbol)
+    {
+        // Check the type itself for Export attributes
+        if (Utils.HasExportAttribute(symbol))
```
AArnott: Will this pick up on `InheritedExportAttribute` from base classes?

AArnott (commented):
Questions about `InheritedExportAttribute`. Ideally you can support that as well. But if you'd rather not, let's just be sure it results in too _few_ diagnostics rather than false positives.

src/Microsoft.VisualStudio.Composition.Analyzers/Utils.cs
drewnoakes: `IsAttributeOfType` walks up the base type chain, so `InheritedExportAttribute` would be considered a match.

src/Microsoft.VisualStudio.Composition.Analyzers/Utils.cs
drewnoakes: Yes.

AArnott:
@drewnoakes do you have a preference to merge vs. squash?

drewnoakes:
Are packages created automatically? Once a version is available, I can try it in a few repos I own and give it a bit more validation.

drewnoakes:
Did the optprof run work? I'm not seeing a new analyzers package yet.

drewnoakes:
Was there any progress on the OptProf run unblocking a new package version? If there are still build failures, I'll take a look (though I'm not sure where to find them).

PR #605, opened 2025-08-05
Support generic type parameters in TypeMismatchAnalyzer

aphistra:
Created some tests and a fix for the case of a generic type, as it looks like this should be valid and not throw a warning.


AArnott (approved):
Looks great!

PR #615, opened 2025-08-30
Migrate from static markdown docs to docfx

AArnott:
These docs are far more presentable and can link directly to API docs and can source sample code from compiled code.


AArnott:
The docfx validate step is failing because this PR includes links to the new docs which won't be published until a few minutes _after_ this PR merges.

PR #698, opened 2026-04-08
Reduce deserialization allocations

davkean:


src/Microsoft.VisualStudio.Composition/Configuration/SerializationContextBase.cs
```diff
@@ -751,6 +758,41 @@ protected Array ReadArray(BinaryReader reader, Func<object?> itemReader, Type el
                     throw new NotSupportedException();
                 }
 
+                // Use typed fast paths for common element types to avoid
+                // the overhead of Array.CreateInstance + Array.SetValue (reflection + boxing).
```
davkean: What boxing is this referring to?

src/Microsoft.VisualStudio.Composition/LazyMetadataWrapper.cs
```diff
@@ -44,6 +46,41 @@ internal LazyMetadataWrapper(ImmutableDictionary<string, object?> metadata, Dire
             this.resolver = resolver;
         }
 
+        internal LazyMetadataWrapper(ImmutableDictionary<string, object?> metadata, Direction direction, Resolver resolver)
```
davkean: Why is this constructor still needed?

src/Microsoft.VisualStudio.Composition/LazyMetadataWrapper.cs
```diff
@@ -44,6 +46,41 @@ internal LazyMetadataWrapper(ImmutableDictionary<string, object?> metadata, Dire
...
+        /// <summary>
+        /// Returns a cached empty metadata wrapper for the specified resolver.
+        /// </summary>
+        internal static LazyMetadataWrapper EmptyMetadata(Resolver resolver) => new LazyMetadataWrapper(EmptyDictionary, Direction.ToOriginalValue, resolver);
+
+        /// <summary>
+        /// Creates a <see cref="Dictionary{TKey, TValue}"/> from an <see cref="IReadOnlyDictionary{TKey, TValue}"/>.
+        /// Avoids a copy if the source is already a <see cref="Dictionary{TKey, TValue}"/> or <see cref="IDictionary{TKey, TValue}"/>.
+        /// </summary>
+        internal static Dictionary<string, object?> ToDictionary(IReadOnlyDictionary<string, object?> source)
```
davkean: Can't we just strongly type this and fix up the callers?

src/Microsoft.VisualStudio.Composition/LazyMetadataWrapper.cs
```diff
@@ -44,6 +46,41 @@ internal LazyMetadataWrapper(ImmutableDictionary<string, object?> metadata, Dire
...
 
+        internal LazyMetadataWrapper(ImmutableDictionary<string, object?> metadata, Direction direction, Resolver resolver)
+            : this(new Dictionary<string, object?>(metadata), direction, resolver)
+        {
+        }
+
+        /// <summary>
+        /// Returns a cached empty metadata wrapper for the specified resolver.
+        /// </summary>
+        internal static LazyMetadataWrapper EmptyMetadata(Resolver resolver) => new LazyMetadataWrapper(EmptyDictionary, Direction.ToOriginalValue, resolver);
```
davkean: What does a empty metadata need with a resolver?

src/Microsoft.VisualStudio.Composition/Configuration/SerializationContextBase.cs
```diff
@@ -815,32 +857,29 @@ protected void Write(IReadOnlyDictionary<string, object?> metadata)
...
-                        object? value = this.ReadObject();
-                        builder.Add(key, value);
-                    }
+                    return LazyMetadataWrapper.EmptyMetadata(this.Resolver);
+                }
+
+                // Reuse a dictionary instance to save on GC pressure.
+                // We take ownership of the dictionary and hand it to LazyMetadataWrapper,
+                // then allocate a new one for the next call.
+                var dictionary = this.reusableMetadataDictionary ?? new Dictionary<string, object?>((int)count);
```
davkean: I don't get this, can you explain it? I don't understand how this saves much or really anything? Wouldn't it be better to always new a dictionary for the metadata using the capacity known at creation time?

src/Microsoft.VisualStudio.Composition/LazyMetadataWrapper.cs
```diff
@@ -265,13 +314,7 @@ protected virtual LazyMetadataWrapper Clone(LazyMetadataWrapper oldVersion, IRea
                 return null;
             }
 
-            value = this.SubstituteValueIfRequired(value);
-
-            // Update our metadata dictionary with the substitution to avoid
-            // the translation costs next time.
-            this.underlyingMetadata = this.underlyingMetadata.SetItem(key, value);
```
davkean, line 272: Why is this no longer applicable?

src/Microsoft.VisualStudio.Composition/RuntimeComposition.cs
```diff
@@ -321,13 +321,24 @@ public bool Equals(RuntimePart? other)
...
+                internal Func<AssemblyName, Func<object?>, object, object>? LazyFactory;
+                internal ParameterInfo? ImportingParameter;
+                internal MemberInfo? ImportingMember;
+                internal volatile bool IsMetadataTypeInitialized;
+                internal Type? MetadataType;
+            }
+
+            private volatile LazyState? lazyState;
+
+            private LazyState GetLazyState() => this.lazyState ?? (this.lazyState = new LazyState());
```
davkean: What do we save by caching these values instead of just calculating them on the fly?

src/Microsoft.VisualStudio.Composition/Configuration/SerializationContextBase.cs
```diff
@@ -801,7 +841,13 @@ protected void Write(IReadOnlyDictionary<string, object?> metadata)
                 // implicitly resolving TypeRefs to Types which is undesirable.
                 metadata = LazyMetadataWrapper.TryUnwrap(metadata);
 
-                serializedMetadata = new LazyMetadataWrapper(metadata.ToImmutableDictionary(), LazyMetadataWrapper.Direction.ToSubstitutedValue, this.Resolver);
+                // Both Dictionary and ImmutableDictionary implement IDictionary,
+                // which Dictionary's copy constructor accepts.
+                var metadataCopy = metadata is IDictionary<string, object?> dict
```
davkean: Is this guaranateed to be Dictionary<string, object?>? This also does not copy the metadata on the second path.

src/Microsoft.VisualStudio.Composition/LazyMetadataWrapper.cs
```diff
@@ -21,20 +21,23 @@ internal class LazyMetadataWrapper : ExportProvider.IMetadataDictionary
             typeof(string).GetTypeInfo().Assembly,
         });
 
+        private static readonly Dictionary<string, object?> EmptyDictionary = new Dictionary<string, object?>();
```
davkean: You have 2 fields and one property for empty metadata, just make it into 1 static field.

src/Microsoft.VisualStudio.Composition/LazyMetadataWrapper.cs
```diff
@@ -253,7 +267,21 @@ internal static bool TryGetLoadSafeValueTypeRef(IReadOnlyDictionary<string, obje
 
         protected virtual LazyMetadataWrapper Clone(LazyMetadataWrapper oldVersion, IReadOnlyDictionary<string, object?> newMetadata)
```
davkean, line 254: Isn't this always Dictionary?

src/Microsoft.VisualStudio.Composition/Configuration/SerializationContextBase.cs
```diff
@@ -59,6 +58,14 @@ internal abstract partial class SerializationContextBase : IDisposable
         private static readonly object BoxedTrue = true;
         private static readonly object BoxedFalse = false;
 
+        private static readonly IReadOnlyDictionary<string, object?> EmptyMetadata = new Dictionary<string, object?>();
```
AArnott: Make absolutely sure it's read-only and empty, and even avoid allocating the dictionary.

```suggestion
        private static readonly IReadOnlyDictionary<string, object?> EmptyMetadata = ImmutableDictionary<string, object?>.Empty;
```

src/Microsoft.VisualStudio.Composition/Configuration/SerializationContextBase.cs
```diff
@@ -801,7 +843,14 @@ protected void Write(IReadOnlyDictionary<string, object?> metadata)
                 // implicitly resolving TypeRefs to Types which is undesirable.
                 metadata = LazyMetadataWrapper.TryUnwrap(metadata);
 
-                serializedMetadata = new LazyMetadataWrapper(metadata.ToImmutableDictionary(), LazyMetadataWrapper.Direction.ToSubstitutedValue, this.Resolver);
+                // Dictionary has no constructor accepting IReadOnlyDictionary, so enumerate manually.
+                var metadataCopy = new Dictionary<string, object?>(metadata.Count);
+                foreach (var entry in metadata)
+                {
+                    metadataCopy.Add(entry.Key, entry.Value);
+                }
```
AArnott: No constructor, perhaps. But you can use this extension method, can't you?

```suggestion
                Dictionary<string, object?> metadataCopy = metadata.ToDictionary();
```

src/Microsoft.VisualStudio.Composition/LazyMetadataWrapper.cs
```diff
@@ -253,7 +253,14 @@ internal static bool TryGetLoadSafeValueTypeRef(IReadOnlyDictionary<string, obje
...
         {
-            return new LazyMetadataWrapper(newMetadata.ToImmutableDictionary(), oldVersion.direction, this.resolver);
+            // Dictionary has no constructor accepting IReadOnlyDictionary, so enumerate manually.
+            var dict = new Dictionary<string, object?>(newMetadata.Count);
+            foreach (var kvp in newMetadata)
+            {
+                dict.Add(kvp.Key, kvp.Value);
+            }
+
+            return new LazyMetadataWrapper(dict, oldVersion.direction, this.resolver);
```
AArnott:
```suggestion
            return new LazyMetadataWrapper(newMetadata.ToDictionary(), oldVersion.direction, this.resolver);
```

src/Microsoft.VisualStudio.Composition/LazyMetadataWrapper.cs
AArnott: Who are you asking? I see a code comment below that seems to justify it.

src/Microsoft.VisualStudio.Composition/RuntimeComposition.cs
```diff
@@ -321,13 +321,21 @@ public bool Equals(RuntimePart? other)
...
-            private ParameterInfo? importingParameter;
-            private MemberInfo? importingMember;
-            private volatile bool isMetadataTypeInitialized;
-            private Type? metadataType;
+            /// <summary>
+            /// Caches for resolved reflection objects and the lazy factory delegate.
+            /// These involve CLR metadata resolution or reflection emit, so they are worth caching.
+            /// Allocated on first access to avoid paying the cost during composition cache loading.
+            /// </summary>
+            private sealed class ResolvedMembers
```
AArnott: Wouldn't a nullable struct be even better?

src/Microsoft.VisualStudio.Composition/RuntimeComposition.cs
```diff
@@ -321,13 +321,21 @@ public bool Equals(RuntimePart? other)
...
-            private Func<AssemblyName, Func<object?>, object, object>? lazyFactory;
-            private ParameterInfo? importingParameter;
-            private MemberInfo? importingMember;
-            private volatile bool isMetadataTypeInitialized;
-            private Type? metadataType;
+            /// <summary>
+            /// Caches for resolved reflection objects and the lazy factory delegate.
+            /// These involve CLR metadata resolution or reflection emit, so they are worth caching.
+            /// Allocated on first access to avoid paying the cost during composition cache loading.
+            /// </summary>
```
AArnott: We were _already_ caching these. Why create a nested type? Was it to shrink the size of `RuntimeImport` objects? Particularly ones that never get accessed?

AArnott (approved):

src/Microsoft.VisualStudio.Composition/LazyMetadataWrapper.cs
davkean: It added this comment after I asked this. :) I could not get it to resolve the question.

src/Microsoft.VisualStudio.Composition/RuntimeComposition.cs
davkean: Yes to reduce the size of RuntimeImport so we only allocate when the data is accessed.

PR #717, opened 2026-05-29
Support generic imports.

### LINKED ISSUE #457: Support generic imports (GenericHost/Options DI)
We've had [Generic Host](https://learn.microsoft.com/en-us/dotnet/core/extensions/generic-host?tabs=appbuilder) for a while now and I've been looking into using vs-mef as DI instead of the inbox implementation, to integrate better with existing codebases which already are MEF based. I've made good progress mapping the builder abstractions and constructing a dynamic vs-mef catalog from ServiceDescriptors during startup, but when I got to the options infrastructure of .NET Core I'm running into the problem that they adopted generic imports, something vs-mef doesn't support yet.

An example for exports added to the builder is [here](https://github.com/dotnet/runtime/blob/a79c62ddc8089cf2879ed36eac9aa333b32bde5f/src/libraries/Microsoft.Extensions.Options/src/OptionsServiceCollectionExtensions.cs#L26):
```csharp
services.TryAdd(ServiceDescriptor.Singleton(typeof(IOptions<>), typeof(UnnamedOptionsManager<>)));
services.TryAdd(ServiceDescriptor.Scoped(typeof(IOptionsSnapshot<>), typeof(OptionsManager<>)));
services.TryAdd(ServiceDescriptor.Singleton(typeof(IOptionsMonitor<>), typeof(OptionsMonitor<>)));
services.TryAdd(ServiceDescriptor.Transient(typeof(IOptionsFactory<>), typeof(OptionsFactory<>)));
services.TryAdd(ServiceDescriptor.Singleton(typeof(IOptionsMonitorCache<>), typeof(OptionsCache<>)));
```

So .NET is creating generic exports, which is fine, vs-mef supports that. However when we look ad e.g. `UnnamedOptionsManager<TOptions>` we can [see its imports](https://github.com/dotnet/runtime/blob/a79c62ddc8089cf2879ed36eac9aa333b32bde5f/src/libraries/Microsoft.Extensions.Options/src/UnnamedOptionsManager.cs#L17):
```csharp
public UnnamedOptionsManager(IOptionsFactory<TOptions> factory)
```

i.e. the import depends on the generic parameter, which is something vs-mef does not support because the TypeRef infrastructure has no strategy to deal with it and just throws.

Conceptually I believe it should be fine if a generic export receives an input parametrized by its own generic parameter? So the question is if there is any desire to have this support added to vs-mef.

weltkante:
Fixes #457

As mentioned, mostly AI generated. The approach makes sense to me. I left the original generated comments intact during cleanup since they might help understanding during review, if further cleanup or other refactoring is desired let me know and I'll work on that. If the approach isn't desireable thats fine too, I just wanted to get this is out of my backlog ;)


test/Microsoft.VisualStudio.Composition.Tests/Reflection/TypeRefTests.cs
```diff
@@ -77,7 +77,7 @@ public void ThrowTypeLoadExceptionWhenArgIsNull()
             var typeRef = TypeRef.Get(TestUtilities.Resolver, assemblyIdentity, 0x02000001, typeof(Dictionary<,>).FullName!, TypeRefFlags.None, 0, new[] { typeRefNullableArgument }.ToImmutableArray(), false, new[] { typeRefNullableArgument }.ToImmutableArray(), null);
 
             var actualException = Assert.Throws<TypeLoadException>(() => typeRef.Resolve());
-            Assert.Contains("Could not load type 'Fake assembly", actualException.Message);
+            Assert.Contains("Fake assembly", actualException.Message);
```
weltkante, line 80: That one was failing for me. Unrelated to the changes, probably a localized error message or something like that

weltkante:
@microsoft-github-policy-service agree

test/Microsoft.VisualStudio.Composition.Tests/GenericImportTests.cs
```diff
@@ -95,5 +95,196 @@ public class PartThatImportsEnumerableOfT<T>
...
+        public class ParameterizedGenericImport_OptionsManager_Lazy<TOptions>
+        {
+            [ImportingConstructor]
+            [MefV1.ImportingConstructor]
+            public ParameterizedGenericImport_OptionsManager_Lazy(Lazy<IParameterizedGenericImport_OptionsFactory<TOptions>> factory)
+            {
+                this.Factory = factory;
+            }
+
+            public Lazy<IParameterizedGenericImport_OptionsFactory<TOptions>> Factory { get; }
```
AArnott, line 173: You have a good set of tests here. But I don't see any direct imports of `T` -- I only see `T` used as a generic type argument in an import. If you expect/want that to work, let's get tests for that added. If it doesn't work and you're not interested in pursuing that, that's OK. 
Either way, we'll want to update the docs to describe what works and what doesn't.

src/Microsoft.VisualStudio.Composition/ReflectionHelpers.cs
```diff
@@ -645,7 +645,40 @@ internal static Type GetContractTypeForDelegate(MethodInfo method)
...
+                if (!parameterTypes.Any(t => t.ContainsGenericParameters))
+                {
+                    return closedGeneric.GetConstructor(BindingFlags.Public | BindingFlags.NonPublic | BindingFlags.Instance, Type.DefaultBinder, parameterTypes, Array.Empty<ParameterModifier>());
+                }
+
+                // Slow path: the open generic constructor has parameters whose types contain generic type
+                // parameters (e.g. IFoo<TOptions> on OptionsManager<TOptions>).  Type.GetConstructor cannot
+                // match those against the closed form (IFoo<MyOptions>), so locate the constructor by
+                // matching the open-generic declaring type's constructor list positionally using metadata token.
+                var openGenericType = method.DeclaringType!.IsGenericType && !method.DeclaringType.IsGenericTypeDefinition
```
AArnott: We should probably avoid using the null forgiveness operator where we're not sure it's valid. In this case, the method is handed a `MethodInfo` and we don't know anything else about it, so it might not actually have a declaring type. But we should be able to adjust this expression to gracefully handle a null declaring type I think.

AArnott (changes requested):
Thanks for sharing this. I think it's shaping up.

src/Microsoft.VisualStudio.Composition/ReflectionHelpers.cs
weltkante: I wasn't aware of a scenario where you can have a declaring type without a method, but you're right, there is no harm in checking this.

test/Microsoft.VisualStudio.Composition.Tests/GenericImportTests.cs
weltkante: > I don't see any direct imports of T -- I only see T used as a generic type argument in an import

direct imports of T were out of scope, as discussed back then in github issue it was about generic imports (IOptionsManager<T>-series of interfaces in the generic host path of the framework).

Representing generic imports directly cannot use the trick of using an open generic type and the existing infrastructure, so it'd be a major change of the representation and logic. I guess one could use a different placeholder trick of the same manner, but its not an obvious extension. I'd leave it out.

> If you expect/want that to work, let's get tests for that added.

These failing scenarios of direct imports of generics actually had preexisting tests (GenericImportTests.cs) so I don't think there's anything to add for this case.

> Either way, we'll want to update the docs to describe what works and what doesn't.

I've added a section the the markdown doc about differences between the versions. Let me know if you had something else in mind.

src/Microsoft.VisualStudio.Composition/ReflectionHelpers.cs
AArnott: Methods without declaring types are probably available via the `DynamicMethod` API. Lightweight code gen can produce methods at runtime with no declaring type.
When .NET added the nullable annotations to these APIs, it seemed they made most properties nullable. And while that's annoying for common cases, they did it to be true to the possible/allowable cases.

docfx/docs/mef_library_differences.md
```diff
@@ -27,6 +27,20 @@ VS MEF follows NuGet MEF's model and supports sub-scopes through NuGet MEF attri
...
 
+## Generic imports
+
+An open generic part may use its own type parameter as a *generic type argument* within an import,
+e.g. `OptionsManager<TOptions>` importing `IOptionsFactory<TOptions>`, satisfied by an open generic
+export such as `[Export(typeof(IOptionsFactory<>))]`. VS MEF supports this for the direct import and
+for the `Lazy<T>`, `ExportFactory<T>`, and `ImportMany` (`IEnumerable<T>` / `T[]`) forms.
+
+Importing a type parameter *directly* — where the import type is the bare parameter itself, e.g.
+`[Import] T` — is honored by .NET MEF and NuGet MEF but **not** by VS MEF. Those runtimes resolve
```
AArnott: I find the doc a little confusing. 

> VS MEF supports this for the direct import and for the `Lazy<T>`

This makes it sound like importing `T` ("direct import") and `Lazy<T>` is allowed. But then in the next paragraph you say `T` can't be directly imported.
And then you say nothing about `Lazy<T>` not being importable, but I strongly suspect `Lazy<T>` cannot be imported, since that would have the same ambiguous graph impact as importing `T`, right?

weltkante: yeah, I double checked (to make sure there's no bug hidden) but this just was confusing wording that slipped through my review. Lazy/ExportFactory/IEnumerable/Array are just the expected special-cased variants in addition over the direct import, but of course with the same constraints as the direct import. I'd probably just drop the explicit enumeration of alternatives and leave it with the initial example?

```suggestion
export such as `[Export(typeof(IOptionsFactory<>))]`.

Importing a type parameter *directly* — where the import type is the bare parameter itself, e.g.
`[Import] T` — is honored by .NET MEF and NuGet MEF but **not** by VS MEF. Those runtimes resolve
```

weltkante:
Ah, thats a test failure that slipped through my local tests, due to the way of it being reported. I'll have a look, seems to be a genuine missed case.

weltkante:
ImportMany attributes left some generic parameters which caused issues roundtripping through a serialized container. I've reverted the logic in the idividual AttributedPartDiscovery branches and instead have it now directly in `GetExportTypeIdentityConstraints` and `GetImportMetadataForGenericTypeImport` - looks cleaner too. Also added a test specifically covering this scenario.

src/Microsoft.VisualStudio.Composition/ReflectionHelpers.cs
```diff
@@ -645,7 +645,43 @@ internal static Type GetContractTypeForDelegate(MethodInfo method)
...
+                    }
+
+                    if (ctorIndex >= 0)
+                    {
+                        var closedCtors = closedGeneric.GetConstructors(BindingFlags.Public | BindingFlags.NonPublic | BindingFlags.Instance);
+                        if (ctorIndex < closedCtors.Length)
+                        {
+                            return closedCtors[ctorIndex];
+                        }
+                    }
```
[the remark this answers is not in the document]
weltkante: Interesting change, can't tell which is better, as far as I'm concerned both sound like implementation details that in practice always hold. But the direct token matching for sure is simpler and allows to get rid of the dual loop, so I'd adopt it if there is no reason against it.

weltkante:
Hmm, that copilot comment brought up some cases that weren't handled. Added tests for them. Also the documentation changes I suggested earlier apparently didn't apply, so I reapplied them manually.

src/Microsoft.VisualStudio.Composition/RuntimeExportProviderFactory+RuntimeExportProvider.cs
```diff
@@ -336,11 +419,40 @@ private PartLifecycleTracker GetOrCreateValue(RuntimeComposition.RuntimeImport i
...
+                }
+
+                if (!importingPartTracker.ImportMetadata.TryGetValue(CompositionConstants.GenericParametersMetadataName, out var outerTypeArgsObj) || outerTypeArgsObj is not Type[] outerTypeArgs)
+                {
+                    return importMetadata;
+                }
+
+                Type[] closedTypeArgs = indexes.Select(i => outerTypeArgs[i]).ToArray();
+                return ImmutableDictionary.CreateRange(importMetadata)
+                    .SetItem(CompositionConstants.GenericParametersMetadataName, closedTypeArgs);
```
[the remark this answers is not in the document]
weltkante: Seems to be a false positive and not reproducable. Quote:

> LazyMetadataWrapper.TryGetValue runs SubstituteValueIfRequired with Direction.ToOriginalValue, so reads through the dictionary always hand back Type[]. The substitution is only observable if you deliberately call LazyMetadataWrapper.TryUnwrap first — which is exactly what GetPartConstructedTypeRef does (to avoid forcing type loads), and why that method needs its two-form handling. GetEffectiveImportMetadata doesn't unwrap, so is not Type[] never trips there.

PR #726, opened 2026-06-08
Propagate obsolete metadata view annotations

AArnott:
When a `[MetadataView]` interface is also marked `[Obsolete]`, the source generator was still emitting a generated implementation class that directly referenced the obsolete interface without being obsolete itself. That produced an avoidable warning from generated code and broke the new regression case.

This change teaches the generator to copy `ObsoleteAttribute` from the metadata-view interface onto the generated class, preserving both constructor and named arguments. I also factored the attribute-emission path so the existing default-value handling and the new obsolete handling share the same argument rendering logic.

A regression test now covers an obsolete metadata-view interface and verifies that the generated implementation carries the matching obsolete annotation.


AArnott:
Closing this as superseded by #727. I folded this commit into that branch because the two PRs touched the same generator and tests and would otherwise conflict.

PR #745, opened 2026-07-11
Reduce composition cache allocations

olegtk:
Reduces allocations while loading composition caches by avoiding lazy metadata wrappers when values require no substitution, reusing common boxed values, creating typed arrays directly, and building the contract export index without LINQ grouping or interface enumerators. Metadata substitution and export ordering remain unchanged.

The VS validation Speedometer run measured `Startup.SingleFile` / `0100.Start VS` with no PIT regressions. `CLR_BytesAllocated_Total_devenv` improved by 1,255,856 bytes (0.87%).

Trace comparisons show `LazyMetadataWrapper` allocations from `SerializationContextBase.ReadMetadata` decreased in every measured iteration:

| Iteration | Baseline | This PR | Reduction |
|---|---:|---:|---:|
| PerfMetrics-1 | 1,514,192 B | 0 B | 1,514,192 B |
| PerfMetrics-2 | 1,394,208 B | 109,512 B | 1,284,696 B |
| PerfMetrics-3 | 1,181,776 B | 107,200 B | 1,074,576 B |
| PerfMetrics-4 | 2,403,120 B | 0 B | 2,403,120 B |
| PerfMetrics-5 | 1,503,048 B | 222,392 B | 1,280,656 B |
| Average | 1,599,269 B | 87,821 B | 1,511,448 B |

[PIT results](https://aka.ms/vsengpit?targetBranch=main&targetBuild=12104.16.Build_0bdbc590.260804.105019.765944&runGroup=Speedometer&baselineRunGroup=Speedometer&baselineBranch=main&baselineBuild=12103.328&category=10&since=2026-07-28T07:00:00.000Z&to=2026-08-05T06:59:59.999Z&testMethodFilter=Startup.SingleFile&selectedCounters=Startup.SingleFile%3E0100.Start%20VS%3ECLR_BytesAllocated_Total_devenv)

The existing `Deserialize` benchmark from #760 exercises the complete cache-load path over its synthetic 312-part catalog on .NET Framework 4.8.1:

| | Mean | Allocated |
|---|---:|---:|
| Baseline (`92f6643d`) | 1.859 ms | 580.62 KB |
| This PR | 1.791 ms | 528.04 KB |
| Delta | -3.7% | -9.1% |

### DIFF
--- a/src/Microsoft.VisualStudio.Composition/Configuration/SerializationContextBase.cs
+++ b/src/Microsoft.VisualStudio.Composition/Configuration/SerializationContextBase.cs
@@ -58,6 +58,12 @@ namespace Microsoft.VisualStudio.Composition
 
         private static readonly object BoxedTrue = true;
         private static readonly object BoxedFalse = false;
+        private static readonly object BoxedCreationPolicyAny = CreationPolicy.Any;
+        private static readonly object BoxedCreationPolicyShared = CreationPolicy.Shared;
+        private static readonly object BoxedCreationPolicyNonShared = CreationPolicy.NonShared;
+        private static readonly object BoxedInt32Zero = 0;
+        private static readonly object BoxedInt32One = 1;
+        private static readonly object BoxedInt32NegativeOne = -1;
 
         internal SerializationContextBase(BinaryReader reader, Resolver resolver)
         {
@@ -751,6 +757,39 @@ namespace Microsoft.VisualStudio.Composition
                     throw new NotSupportedException();
                 }
 
+                if (elementType == typeof(object))
+                {
+                    var array = new object?[(int)count];
+                    for (int i = 0; i < array.Length; i++)
+                    {
+                        array[i] = itemReader();
+                    }
+
+                    return array;
+                }
+
+                if (elementType == typeof(string))
+                {
+                    var array = new string?[(int)count];
+                    for (int i = 0; i < array.Length; i++)
+                    {
+                        array[i] = (string?)itemReader();
+                    }
+
+                    return array;
+                }
+
+                if (elementType == typeof(Type))
+                {
+                    var array = new Type?[(int)count];
+                    for (int i = 0; i < array.Length; i++)
+                    {
+                        array[i] = (Type?)itemReader();
+                    }
+
+                    return array;
+                }
+
                 var list = Array.CreateInstance(elementType, (int)count);
                 for (int i = 0; i < list.Length; i++)
                 {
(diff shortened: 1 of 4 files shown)


src/Microsoft.VisualStudio.Composition/RuntimeComposition.cs
```diff
@@ -75,6 +67,33 @@ public IReadOnlyDictionary<TypeRef, RuntimeExport> MetadataViewsAndProviders
...
+        {
+            var exportsByContractName = new Dictionary<string, List<RuntimeExport>>(StringComparer.Ordinal);
+            foreach (RuntimePart part in parts)
+            {
+                for (int i = 0; i < part.Exports.Count; i++)
+                {
+                    RuntimeExport export = part.Exports[i];
+                    if (!exportsByContractName.TryGetValue(export.ContractName, out List<RuntimeExport>? exports))
+                    {
+                        exports = new List<RuntimeExport>(capacity: 1);
```
sandyarmstrong, line 80: TIL that capacity is initially `0`. But should we follow the heuristic that exists in `List<T>`, and set the initial capacity to `4` when we know at least one item will be added? Do we have any idea of the typical length of these lists?

sandyarmstrong (approved):

src/Microsoft.VisualStudio.Composition/RuntimeComposition.cs
AmadeusW: can we set capacity to 1 for concrete types and 4 for interfaces?
Alternatively, we could trace what's the final size of each of the exports lists.

olegtk: The benchmark catalog has 23 singleton contract groups and three multi-export groups (10, 80, and 200 exports). With List growth, capacity 1 allocates fewer backing slots for this mix than 4, so I kept it.

AmadeusW: thanks for checking! capacity 1 makes sense then

AmadeusW (approved):

src/Microsoft.VisualStudio.Composition/Configuration/SerializationContextBase.cs
```diff
@@ -815,32 +854,28 @@ protected void Write(IReadOnlyDictionary<string, object?> metadata)
...
+                {
+                    string? key = this.ReadString();
+                    object? value = this.ReadObject();
+                    requiresValueSubstitution |= value is LazyMetadataWrapper.ISubstitutedValue;
+                    builder.Add(key, value);
                 }
 
-                return new LazyMetadataWrapper(metadata, LazyMetadataWrapper.Direction.ToOriginalValue, this.Resolver);
+                ImmutableDictionary<string, object?> metadata = builder.ToImmutable();
+                builder.Clear();
```
AArnott: The comments that were lost here and where the `builder` local were assigned seem valuable to keep.

AArnott (approved):

src/Microsoft.VisualStudio.Composition/Configuration/SerializationContextBase.cs
olegtk: Restored the deferred-deserialization note and the builder lifecycle comments.

*Cowritten with Copilot (GitHub Copilot)*

davkean (approved):

PR #746, opened 2026-07-14
Reduce metadata view activation overhead

sgreenmsft:
Source-generated metadata views repeatedly discovered and invoked their constructors through reflection each time an export's metadata was accessed.

Cache validated activation plans by metadata and implementation type. Bind a typed construction delegate once for MetadataView implementations while preserving reflective dictionary-constructor support and existing exception behavior.

### DIFF
--- a/src/Microsoft.VisualStudio.Composition/Configuration/MetadataViewImplProxy.cs
+++ b/src/Microsoft.VisualStudio.Composition/Configuration/MetadataViewImplProxy.cs
@@ -4,6 +4,7 @@
 namespace Microsoft.VisualStudio.Composition
 {
     using System;
+    using System.Collections.Concurrent;
     using System.Collections.Generic;
     using System.Collections.Immutable;
     using System.Globalization;
@@ -13,6 +14,10 @@ namespace Microsoft.VisualStudio.Composition
 
     internal class MetadataViewImplProxy : IMetadataViewProvider
     {
+        private static readonly ConcurrentDictionary<Type, ImplementationActivation> ImplementationActivations = new ConcurrentDictionary<Type, ImplementationActivation>();
+        private static readonly ConcurrentDictionary<(Type MetadataType, Type ImplementationType), ImplementationActivation> ExplicitImplementationActivations = new ConcurrentDictionary<(Type MetadataType, Type ImplementationType), ImplementationActivation>();
+        private static readonly MethodInfo CreateMetadataViewMethod = typeof(MetadataViewImplProxy).GetMethod(nameof(CreateMetadataView), BindingFlags.NonPublic | BindingFlags.Static)!;
+
         internal static readonly ComposablePartDefinition PartDefinition =
             Utilities.GetMetadataViewProviderPartDefinition(typeof(MetadataViewImplProxy), 100, Resolver.DefaultInstance);
 
@@ -25,15 +30,14 @@ namespace Microsoft.VisualStudio.Composition
 
         public bool IsMetadataViewSupported(Type metadataType)
         {
-            Type? implementationType = GetMetadataViewImplementationType(metadataType);
-            return implementationType is object && TryGetImplementationActivation(metadataType, implementationType, throwOnInvalidConfiguration: true) is object;
+            return TryGetImplementationActivation(metadataType, throwOnInvalidConfiguration: true) is object;
         }
 
         public object CreateProxy(IReadOnlyDictionary<string, object?> metadata, IReadOnlyDictionary<string, object?> defaultValues, Type metadataViewType)
         {
-            Type? implementationType = GetMetadataViewImplementationType(metadataViewType);
-            Assumes.NotNull(implementationType);
-            return CreateProxy(metadata, defaultValues, metadataViewType, implementationType);
+            var activation = TryGetImplementationActivation(metadataViewType, throwOnInvalidConfiguration: true);
+            Assumes.NotNull(activation);
+            return activation.Value.CreateProxy(metadata, defaultValues, metadataViewType);
         }
 
         internal static bool IsMetadataViewImplementationSupported(Type metadataType, Type implementationType, bool throwOnInvalidConfiguration = true)
@@ -62,8 +66,24 @@ namespace Microsoft.VisualStudio.Composition
         {
             Requires.NotNull(metadataType, nameof(metadataType));
 
+            if (ImplementationActivations.TryGetValue(metadataType, out ImplementationActivation cachedActivation))
+            {
+                return cachedActivation;
+            }
+
             Type? implementationType = GetMetadataViewImplementationType(metadataType);
-            return implementationType is object ? TryGetImplementationActivation(metadataType, implementationType, throwOnInvalidConfiguration) : null;
+            if (implementationType is null)
+            {
+                return null;
+            }
+
+            ImplementationActivation? activation = TryGetImplementationActivation(metadataType, implementationType, throwOnInvalidConfiguration);
+            if (activation.HasValue)
+            {
+                return ImplementationActivations.GetOrAdd(metadataType, activation.Value);
+            }
+
+            return null;
         }
 
         private static Type? GetMetadataViewImplementationType(Type metadataType)
@@ -79,6 +99,23 @@ namespace Microsoft.VisualStudio.Composition
             Requires.NotNull(metadataType, nameof(metadataType));
             Requires.NotNull(implementationType, nameof(implementationType));
 
+            var key = (MetadataType: metadataType, ImplementationType: implementationType);
+            if (ExplicitImplementationActivations.TryGetValue(key, out ImplementationActivation cachedActivation))
+            {
+                return cachedActivation;
+            }
+
+            ImplementationActivation? activation = CreateImplementationActivation(metadataType, implementationType, throwOnInvalidConfiguration);
+            if (activation.HasValue)
+            {
+                return ExplicitImplementationActivations.GetOrAdd(key, activation.Value);
+            }
+
+            return null;
+        }
+
+        private static ImplementationActivation? CreateImplementationActivation(Type metadataType, Type implementationType, bool throwOnInvalidConfiguration)
+        {
             if (!metadataType.IsAssignableFrom(implementationType))
             {
                 if (throwOnInvalidConfiguration)
@@ -109,6 +146,18 @@ namespace Microsoft.VisualStudio.Composition
             }
 
             return null;
+        }
+
+        private static MetadataViewFactory CreateMetadataViewFactory(Type implementationType)
+        {
+            MethodInfo factoryMethod = CreateMetadataViewMethod.MakeGenericMethod(implementationType);
+            return (MetadataViewFactory)factoryMethod.CreateDelegate(typeof(MetadataViewFactory));
+        }
+
+        private static MetadataView CreateMetadataView<T>()
+            where T : MetadataView, new()
+        {
+            return new T();
         }
 
         private static ConstructorInfo? FindDefaultConstructor(TypeInfo implementationType)
@@ -144,17 +193,23 @@ namespace Microsoft.VisualStudio.Composition
                 || parameterType.IsAssignableFrom(typeof(ImmutableDictionary<string, object?>)));
         }
 
+        private delegate MetadataView MetadataViewFactory();
+
         private readonly struct ImplementationActivation
         {
+            private readonly ConstructorInfo? constructor;
+            private readonly Type? constructorParameterType;
+            private readonly MetadataViewFactory? metadataViewFactory;
+
             internal ImplementationActivation(ImplementationKind kind, ConstructorInfo constructor)
             {
                 this.Kind = kind;
-                this.Constructor = constructor;
+                this.constructor = kind == ImplementationKind.LegacyDictionary ? constructor : null;
+                this.constructorParameterType = kind == ImplementationKind.LegacyDictionary ? constructor.GetParameters()[0].ParameterType : null;
+                this.metadataViewFactory = kind == ImplementationKind.MetadataViewBase ? CreateMetadataViewFactory(constructor.DeclaringType!) : null;
             }
 
             internal ImplementationKind Kind { get; }
-
-            internal ConstructorInfo Constructor { get; }
 
             internal object CreateProxy(IReadOnlyDictionary<string, object?> metadata, IReadOnlyDictionary<string, object?> defaultValues, Type metadataViewType)
             {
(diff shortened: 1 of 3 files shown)


src/Microsoft.VisualStudio.Composition/Configuration/MetadataViewImplProxy.cs
```diff
@@ -111,6 +148,18 @@ internal static ImplementationKind GetImplementationKind(Type metadataType)
...
+        {
+            MethodInfo factoryMethod = CreateMetadataViewMethod.MakeGenericMethod(implementationType);
+            return (MetadataViewFactory)factoryMethod.CreateDelegate(typeof(MetadataViewFactory));
+        }
+
+        private static MetadataView CreateMetadataView<T>()
+            where T : MetadataView, new()
+        {
+            return new T();
+        }
```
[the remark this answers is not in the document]
sgreenmsft: Good news -- the MetadataViewFactory delegate throws a TargetInvocationException.

The relevant test passes locally:
PS Q:\src\CPS> dotnet test Q:\src\vs-mef\test\Microsoft.VisualStudio.Composition.Tests\Microsoft.VisualStudio.Composition.Tests.csproj -c Release --no-build --filter "FullyQualifiedName~MetadataViewImplementationWrapsConstructorException"
[xUnit.net 00:00:00.00] xUnit.net VSTest Adapter v3.1.5+1b188a7b0a (64-bit .NET 8.0.28)
[xUnit.net 00:00:00.15]   Discovering: Microsoft.VisualStudio.Composition.Tests
[xUnit.net 00:00:00.42]   Discovered:  Microsoft.VisualStudio.Composition.Tests
[xUnit.net 00:00:00.44]   Starting:    Microsoft.VisualStudio.Composition.Tests
[xUnit.net 00:00:00.49]   Finished:    Microsoft.VisualStudio.Composition.Tests
  Microsoft.VisualStudio.Composition.Tests test net8.0 succeeded (1.6s)
[xUnit.net 00:00:00.00] xUnit.net VSTest Adapter v3.1.5+1b188a7b0a (64-bit .NET Framework 4.8.9324.0)
[xUnit.net 00:00:00.86]   Discovering: Microsoft.VisualStudio.Composition.Tests
[xUnit.net 00:00:01.74]   Discovered:  Microsoft.VisualStudio.Composition.Tests
[xUnit.net 00:00:01.81]   Starting:    Microsoft.VisualStudio.Composition.Tests
[xUnit.net 00:00:01.94]   Finished:    Microsoft.VisualStudio.Composition.Tests
  Microsoft.VisualStudio.Composition.Tests test net472 succeeded (3.5s)

Test summary: total: 2, failed: 0, succeeded: 2, skipped: 0, duration: 4.1s
Build succeeded in 4.7s

ildasm output:
PS Q:\src\CPS> $assembly = 'Q:\src\vs-mef\bin\Microsoft.VisualStudio.Composition\Release\net8.0\Microsoft.VisualStudio.Composition.dll'; & 'C:\Program Files (x86)\Microsoft SDKs\Windows\v10.0A\bin\NETFX 4.8.1 Tools\ildasm.exe' $assembly /text /nobar "/item:Microsoft.VisualStudio.Composition.MetadataViewImplProxy::CreateMetadataView" | Out-String                                                    

//  Microsoft (R) .NET Framework IL Disassembler.  Version 4.8.3928.0
//  Copyright (c) Microsoft Corporation.  All rights reserved.


// warning : THIS IS A PARTIAL DISASSEMBLY, NOT SUITABLE FOR RE-ASSEMBLING



.class private auto ansi beforefieldinit Microsoft.VisualStudio.Composition.MetadataViewImplProxy
       extends [System.Runtime]System.Object
       implements Microsoft.VisualStudio.Composition.IMetadataViewProvider
{
  .custom instance void [System.Runtime]System.Runtime.CompilerServices.NullableContextAttribute::.ctor(uint8) = ( 01 00 01 00 00 ) 
  .custom instance void [System.Runtime]System.Runtime.CompilerServices.NullableAttribute::.ctor(uint8) = ( 01 00 00 00 00 ) 
  .interfaceimpl type Microsoft.VisualStudio.Composition.IMetadataViewProvider
  .custom instance void [System.Runtime]System.Runtime.CompilerServices.NullableAttribute::.ctor(uint8) = ( 01 00 00 00 00 ) 
  .method private hidebysig static class Microsoft.VisualStudio.Composition.MetadataView 
          CreateMetadataView<.ctor (Microsoft.VisualStudio.Composition.MetadataView) T>() cil managed
  {
    .param type T 
    .custom instance void [System.Runtime]System.Runtime.CompilerServices.NullableAttribute::.ctor(uint8) = ( 01 00 00 00 00 ) 
    // Code size       11 (0xb)
    .maxstack  8
    IL_0000:  call       !!0 [System.Runtime]System.Activator::CreateInstance<!!0>()
    IL_0005:  box        !!T
    IL_000a:  ret
  } // end of method MetadataViewImplProxy::CreateMetadataView

} // end of class Microsoft.VisualStudio.Composition.MetadataViewImplProxy

// *********** DISASSEMBLY COMPLETE ***********************

src/Microsoft.VisualStudio.Composition/Configuration/MetadataViewImplProxy.cs
```diff
@@ -62,8 +66,24 @@ internal static ImplementationKind GetImplementationKind(Type metadataType)
...
+            if (ImplementationActivations.TryGetValue(metadataType, out ImplementationActivation cachedActivation))
+            {
+                return cachedActivation;
+            }
+
             Type? implementationType = GetMetadataViewImplementationType(metadataType);
-            return implementationType is object ? TryGetImplementationActivation(metadataType, implementationType, throwOnInvalidConfiguration) : null;
+            if (implementationType is null)
+            {
+                return null;
```
AArnott: 🏎️ Every time we `return null` we don't cache how we got to this point, so on repeated requests with the same input, we'll repeatedly call `GetMetadataViewImplementationType` which will invoke reflection to `GetFirstAttribute` before determining there's no implementation there.
It seems like we should cache null values in the concurrent dictionary as well.

src/Microsoft.VisualStudio.Composition/Configuration/MetadataViewImplProxy.cs
```diff
@@ -13,6 +14,10 @@ namespace Microsoft.VisualStudio.Composition
 
     internal class MetadataViewImplProxy : IMetadataViewProvider
     {
+        private static readonly ConcurrentDictionary<Type, ImplementationActivation> ImplementationActivations = new ConcurrentDictionary<Type, ImplementationActivation>();
+        private static readonly ConcurrentDictionary<(Type MetadataType, Type ImplementationType), ImplementationActivation> ExplicitImplementationActivations = new ConcurrentDictionary<(Type MetadataType, Type ImplementationType), ImplementationActivation>();
```
AArnott: 💅 Use target type creation syntax for shorter, easier-to-read lines.

```suggestion
        private static readonly ConcurrentDictionary<Type, ImplementationActivation> ImplementationActivations = new();
        private static readonly ConcurrentDictionary<(Type MetadataType, Type ImplementationType), ImplementationActivation> ExplicitImplementationActivations = new();
```

src/Microsoft.VisualStudio.Composition/Configuration/MetadataViewImplProxy.cs
```diff
@@ -25,15 +30,14 @@ internal enum ImplementationKind
...
-            return implementationType is object && TryGetImplementationActivation(metadataType, implementationType, throwOnInvalidConfiguration: true) is object;
+            return TryGetImplementationActivation(metadataType, throwOnInvalidConfiguration: true) is object;
         }
 
         public object CreateProxy(IReadOnlyDictionary<string, object?> metadata, IReadOnlyDictionary<string, object?> defaultValues, Type metadataViewType)
         {
-            Type? implementationType = GetMetadataViewImplementationType(metadataViewType);
-            Assumes.NotNull(implementationType);
-            return CreateProxy(metadata, defaultValues, metadataViewType, implementationType);
+            var activation = TryGetImplementationActivation(metadataViewType, throwOnInvalidConfiguration: true);
```
AArnott: 💅 Avoid using `var` when the type isn't obvious from the line of code.

```suggestion
            ImplementationActivation? activation = TryGetImplementationActivation(metadataViewType, throwOnInvalidConfiguration: true);
```

src/Microsoft.VisualStudio.Composition/Configuration/MetadataViewImplProxy.cs
```diff
@@ -164,22 +219,21 @@ internal object CreateProxy(IReadOnlyDictionary<string, object?> metadata, IRead
 
                 return this.Kind switch
                 {
-                    ImplementationKind.MetadataViewBase => InitializeMetadataView(metadata, defaultValues, this.Constructor),
-                    ImplementationKind.LegacyDictionary => InvokeDictionaryConstructor(metadata, this.Constructor),
+                    ImplementationKind.MetadataViewBase => InitializeMetadataView(metadata, defaultValues, this.metadataViewFactory!),
+                    ImplementationKind.LegacyDictionary => InvokeDictionaryConstructor(metadata, this.constructor!, this.constructorParameterType!),
```
AArnott: 💡 You can avoid all the null forgiveness operators by replacing this switch expression to one that checks the fields themselves against null and using the one that is non-null.
For the `this.constructor, this.constructorParameterType` pair, you could store both of those in a nullable value tuple field so that you know either neither or both are null.

AArnott (changes requested):
Thanks for this. A few touch-ups and a perf improvement and I think we'll be ready to go.

test/Microsoft.VisualStudio.Composition.Tests/MetadataViewImplementationTests.cs
```diff
@@ -355,6 +384,20 @@ private static object CreateProxy(IReadOnlyDictionary<string, object?> metadata,
...
+                "TryGetImplementationActivation",
+                BindingFlags.Static | BindingFlags.NonPublic,
+                binder: null,
+                new[] { typeof(Type), typeof(bool) },
+                modifiers: null)!;
+            object activation = getActivationMethod.Invoke(null, new object[] { metadataViewType, true })!;
+            FieldInfo factoryField = activation.GetType().GetField("metadataViewFactory", BindingFlags.Instance | BindingFlags.NonPublic)!;
+            return factoryField.GetValue(activation)!;
+        }
+
```
AArnott: 👮 Using reflection to verify an implementation detail is counter to the intent of avoiding `InternalsVisibleTo`. We don't want brittle tests.
I realize that the very next method below does a similar thing... I don't recall why I let that through -- I might have missed it.
I'd rather you delete the test that needs this than that we keep this.

src/Microsoft.VisualStudio.Composition/Configuration/MetadataViewImplProxy.cs
```diff
@@ -13,6 +14,10 @@ namespace Microsoft.VisualStudio.Composition
 
     internal class MetadataViewImplProxy : IMetadataViewProvider
     {
+        private static readonly ConcurrentDictionary<Type, ImplementationActivationCacheEntry> ImplementationActivations = new();
+        private static readonly ConcurrentDictionary<(Type MetadataType, Type ImplementationType), ImplementationActivation> ExplicitImplementationActivations = new();
+        private static readonly MethodInfo CreateMetadataViewMethod = typeof(MetadataViewImplProxy).GetMethod(nameof(CreateMetadataView), BindingFlags.NonPublic | BindingFlags.Static)!;
```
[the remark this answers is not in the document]
AArnott: I considered this as well. But weak references have a perf cost of their own. So until we have collectible ALCs as a target scenario, this is the better tradeoff.

src/Microsoft.VisualStudio.Composition/Configuration/MetadataViewImplProxy.cs
```diff
@@ -195,18 +197,33 @@ private static bool IsSupportedDictionaryParameterType(Type parameterType)
...
 
+        private readonly struct ImplementationActivationCacheEntry
+        {
+            internal ImplementationActivationCacheEntry(ImplementationActivation activation)
+            {
+                this.Activation = activation;
+            }
+
+            internal ImplementationActivation? Activation { get; }
+        }
```
AArnott: We did we need a whole new type just to store one nullable value? ConcurrentDictionary could have stored `ImplementationActivation?` as the TValue, couldn't it?

AArnott (changes requested):
Better. One new comment.

AArnott (approved):

src/Microsoft.VisualStudio.Composition/Configuration/MetadataViewImplProxy.cs
```diff
@@ -94,23 +151,49 @@ internal static ImplementationKind GetImplementationKind(Type metadataType)
             ConstructorInfo? ctor = FindDefaultConstructor(implementationTypeInfo);
             if (ctor != null && typeof(MetadataView).GetTypeInfo().IsAssignableFrom(implementationTypeInfo))
             {
-                return new ImplementationActivation(ImplementationKind.MetadataViewBase, ctor);
+                return new ImplementationActivation(CreateMetadataViewFactory(implementationType));
             }
```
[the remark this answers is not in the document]
sgreenmsft: So I haven't checked the full history of handling of abstract types, but when given an abstract type in IsMetadataViewSupported the code currently in main returns true. If given to CreateProxy, it'll throw a MemberAccessException. I've updated the PR to add a check for abstract types and throw an InvalidOperationException with a helpful message.

AArnott (approved):

src/Microsoft.VisualStudio.Composition/Configuration/MetadataViewImplProxy.cs
```diff
@@ -79,38 +102,109 @@ internal static ImplementationKind GetImplementationKind(Type metadataType)
...
+        {
+            MethodInfo factoryMethod = CreateMetadataViewMethod.MakeGenericMethod(implementationType);
+            return (MetadataViewFactory)factoryMethod.CreateDelegate(typeof(MetadataViewFactory));
+        }
+
+        private static MetadataView CreateMetadataView<T>()
+            where T : MetadataView, new()
+        {
+            return new T();
+        }
```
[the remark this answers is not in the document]
sgreenmsft: This is incorrect. For details please refer to my reply to the previous instance of this comment.

PR #760, opened 2026-07-24
Reduce ComposableCatalog.GetExports allocations (-35% composition) + add BenchmarkDotNet harness

karpinsn:
Optimizes the export-filtering hot path in `ComposableCatalog.GetExports`, which runs once
per import site during `CompositionConfiguration.Create`:

- Replaces `ExportConstraints.All(c => c.IsSatisfiedBy(export))` with a manual loop,
  eliminating a per-candidate closure + `Func` delegate + boxed enumerator.
- Returns a plain array instead of building an `ImmutableList` (no per-element tree nodes).

Adds a `Microsoft.VisualStudio.Composition.Benchmarks` BenchmarkDotNet project (with a
DiagnosticsHub `[CPUUsageDiagnoser]` for `.diagsession` traces) to measure it.

Composition benchmark (BenchmarkDotNet, ~312-part catalog):

| | Mean | Allocated |
|---|---|---|
| Before | 17.60 ms | 9.02 MB |
| After | 13.50 ms | 5.86 MB |
| Delta | -23.3% | -35.0% |

All Composition (2148) and Operations (662) tests pass.


perf/Microsoft.VisualStudio.Composition.Benchmarks/Microsoft.VisualStudio.Composition.Benchmarks.csproj
```diff
@@ -0,0 +1,31 @@
...
+
+  <PropertyGroup>
+    <OutputType>Exe</OutputType>
+    <TargetFramework>net8.0</TargetFramework>
+    <Nullable>disable</Nullable>
+    <ImplicitUsings>disable</ImplicitUsings>
+    <GenerateDocumentationFile>false</GenerateDocumentationFile>
+    <EnforceCodeStyleInBuild>false</EnforceCodeStyleInBuild>
+    <EnableNETAnalyzers>false</EnableNETAnalyzers>
+    <RunAnalyzers>false</RunAnalyzers>
```
AArnott: I tend to want all of these things enabled.

perf/Microsoft.VisualStudio.Composition.Benchmarks/Microsoft.VisualStudio.Composition.Benchmarks.csproj
```diff
@@ -0,0 +1,31 @@
...
+    <IsPackable>false</IsPackable>
+    <EnablePackageValidation>false</EnablePackageValidation>
```
AArnott: I don't think you need to turn off package validation when you've already turned off packing.

perf/Microsoft.VisualStudio.Composition.Benchmarks/Microsoft.VisualStudio.Composition.Benchmarks.csproj
```diff
@@ -0,0 +1,31 @@
...
+  <PropertyGroup>
+    <OutputType>Exe</OutputType>
+    <TargetFramework>net8.0</TargetFramework>
+    <Nullable>disable</Nullable>
+    <ImplicitUsings>disable</ImplicitUsings>
+    <GenerateDocumentationFile>false</GenerateDocumentationFile>
+    <EnforceCodeStyleInBuild>false</EnforceCodeStyleInBuild>
+    <EnableNETAnalyzers>false</EnableNETAnalyzers>
+    <RunAnalyzers>false</RunAnalyzers>
+    <IsPackable>false</IsPackable>
```
AArnott: I tend to put benchmarks in the `test` subfolder. And if you do that, packing is turned off automatically.

perf/Microsoft.VisualStudio.Composition.Benchmarks/Microsoft.VisualStudio.Composition.Benchmarks.csproj
```diff
@@ -0,0 +1,31 @@
+<Project Sdk="Microsoft.NET.Sdk">
+
+  <PropertyGroup>
+    <OutputType>Exe</OutputType>
+    <TargetFramework>net8.0</TargetFramework>
```
AArnott: I would focus on .NET 10 for perf benchmarks. Or multi-target and include net472. But focusing on _only_ .NET 8 is not so relevant when it's only a few short months away from EOL.

perf/Microsoft.VisualStudio.Composition.Benchmarks/Microsoft.VisualStudio.Composition.Benchmarks.csproj
```diff
@@ -0,0 +1,31 @@
...
+    <GenerateDocumentationFile>false</GenerateDocumentationFile>
+    <EnforceCodeStyleInBuild>false</EnforceCodeStyleInBuild>
+    <EnableNETAnalyzers>false</EnableNETAnalyzers>
+    <RunAnalyzers>false</RunAnalyzers>
+    <IsPackable>false</IsPackable>
+    <EnablePackageValidation>false</EnablePackageValidation>
+    <ProduceReferenceAssembly>false</ProduceReferenceAssembly>
+    <!-- Let BenchmarkDotNet pick its own transitive versions; the repo's central pins force a downgrade. -->
+    <CentralPackageTransitivePinningEnabled>false</CentralPackageTransitivePinningEnabled>
+    <NoWarn>$(NoWarn);CS1591;SA0001;CS8032;NU1701;MSB3277</NoWarn>
```
AArnott: At least some of these suppressions are already taken care of in the `test` directory. And any other warnings that surface should perhaps be dealt with.

perf/Microsoft.VisualStudio.Composition.Benchmarks/Microsoft.VisualStudio.Composition.Benchmarks.csproj
```diff
@@ -0,0 +1,31 @@
...
+    <Optimize>true</Optimize>
+    <ServerGarbageCollection>false</ServerGarbageCollection>
+    <ConcurrentGarbageCollection>false</ConcurrentGarbageCollection>
+  </PropertyGroup>
+
+  <ItemGroup>
+    <ProjectReference Include="..\..\src\Microsoft.VisualStudio.Composition\Microsoft.VisualStudio.Composition.csproj" />
```
AArnott: Please reserve each item group to just one item type instead of mixing them.

perf/Microsoft.VisualStudio.Composition.Benchmarks/Program.cs
```diff
@@ -0,0 +1,13 @@
+// Copyright (c) Microsoft Corporation. All rights reserved.
+// Licensed under the MIT license. See LICENSE file in the project root for full license information.
+
+namespace Microsoft.VisualStudio.Composition.Benchmarks
```
AArnott: Please use `namespace` statements instead of blocks for all new files.
And while you're at it, I'd say drop the class and main method, leaving only the one statement as the only line of code (other than the `using`) in this file.

src/Microsoft.VisualStudio.Composition/ComposableCatalog.cs
```diff
@@ -239,11 +239,44 @@ from export in openGenericExports
...
+            foreach (ExportDefinitionBinding export in exports)
+            {
+                if (IsExportSatisfiedByAllConstraints(constraints, export.ExportDefinition))
+                {
+                    filteredExports.Add(export);
+                }
+            }
+
+            // Return an array, not an ImmutableList, to avoid its per-element tree-node allocations.
+            return filteredExports.ToArray();
```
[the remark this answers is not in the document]
AArnott: Presizing is a good idea. I don't remember how likely having _no_ export constraints is. I vaguely recall that most imports have at least one or two constraints (e.g. assignability). So the second suggestion is fine, maybe not too effective, but certainly not harmful so we should probably do both ideas.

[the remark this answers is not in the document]
karpinsn: Thanks — addressed in  7fed5c37 .
- No-constraint fast path: when  ExportConstraints.Count == 0  we return  exports.ToArray()  directly, skipping the filter loop and intermediate  List<> .
- Pre-size: to be clear about the sizing concern —  exports  is a single contract's bucket from  exportsByContract  (exports sharing the imported contract name), not the whole catalog; typically 1, and it's the upper bound on the result, so the pre-size stays small.
- Regression guard: added  GetExportsUnconstrained / GetExportsConstrained  (both resolve the same 200 exports): unconstrained 7.6 µs / 1.61 KB vs constrained 30 µs / 3.23 KB. All 2057 Composition tests pass.

— posted on behalf of the author via GitHub Copilot CLI

perf/Microsoft.VisualStudio.Composition.Benchmarks/CompositionBenchmarks.cs
```diff
@@ -0,0 +1,95 @@
...
+            var loaded = new CachedComposition().LoadRuntimeCompositionAsync(ms, this.resolver).GetAwaiter().GetResult();
+            return loaded.Parts.Count;
+        }
+
+        [Benchmark]
+        public int Runtime()
+        {
+            int count = 0;
+            var exportProvider = this.exportProviderFactory.CreateExportProvider();
+            foreach (var processor in exportProvider.GetExportedValues<IProcessor>())
```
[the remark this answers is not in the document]
karpinsn: Good catch — fixed in  7fed5c37 . The  Runtime  benchmark now wraps the provider in  using var  so instantiated parts don't accumulate across iterations and skew memory numbers.

— posted on behalf of the author via GitHub Copilot CLI

AArnott (changes requested):
Thanks so much for sharing this. I look forward to merging it.

perf/Microsoft.VisualStudio.Composition.Benchmarks/Microsoft.VisualStudio.Composition.Benchmarks.csproj
karpinsn: Fixed

perf/Microsoft.VisualStudio.Composition.Benchmarks/Microsoft.VisualStudio.Composition.Benchmarks.csproj
karpinsn: Fixed

perf/Microsoft.VisualStudio.Composition.Benchmarks/Microsoft.VisualStudio.Composition.Benchmarks.csproj
karpinsn: Fixed

perf/Microsoft.VisualStudio.Composition.Benchmarks/Microsoft.VisualStudio.Composition.Benchmarks.csproj
karpinsn: I'll target .NET 10. The Profiler diagnosers support multi-targeting but I find it just takes longer to run benchmarks and I try to stay away from framework specific optimizations. Happy to change it if you want though.

perf/Microsoft.VisualStudio.Composition.Benchmarks/Microsoft.VisualStudio.Composition.Benchmarks.csproj
karpinsn: Fixed

perf/Microsoft.VisualStudio.Composition.Benchmarks/Microsoft.VisualStudio.Composition.Benchmarks.csproj
karpinsn: Fixed

perf/Microsoft.VisualStudio.Composition.Benchmarks/Program.cs
karpinsn: Fixed.

test/Microsoft.VisualStudio.Composition.Benchmarks/Microsoft.VisualStudio.Composition.Benchmarks.csproj
```diff
@@ -0,0 +1,21 @@
+<Project Sdk="Microsoft.NET.Sdk">
+
+  <PropertyGroup>
+    <OutputType>Exe</OutputType>
+    <TargetFramework>net10.0</TargetFramework>
+    <!-- BenchmarkDotNet must resolve its own transitive versions; the repo's central pins force a downgrade. -->
+    <CentralPackageTransitivePinningEnabled>false</CentralPackageTransitivePinningEnabled>
```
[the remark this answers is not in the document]
karpinsn: Thanks for the review! Addressed the latest feedback and squashed the "addressing feedback" commits into one — the branch is now 3 commits (harness, the  GetExports  optimization, and a single "Address review feedback" commit), so I force-pushed.

 GetExports  no-constraint fast path — good catch, you're right.  exports  is already an  ImmutableList<ExportDefinitionBinding>  and the method returns  IReadOnlyList<…> , so returning it directly is safe (it's immutable — no aliasing concern) and skips the copy entirely. The  GetExportsUnconstrained  benchmark drops from 1.61 KB → 24 B. Changed  return exports.ToArray();  →  return exports; .

 CentralPackageTransitivePinningEnabled  — removed it, and it turns out both the per-project override and the  Directory.Packages.props  autogenerated-project condition are now unnecessary, so I dropped both. The override was only needed on net8, where the central  System.Collections.Immutable 8.0.0  pin conflicted with BenchmarkDotNet's transitive Roslyn (which wants ≥ 10.0.1). After moving to net10, package pruning lets the framework supply those assemblies at 10.x, so central pinning no longer triggers a downgrade. Verified the benchmark still builds and runs (autogenerated project included) with no CPM special-casing anywhere.

Builds clean on net10.0 with analyzers/code-style/docs enabled, and all 2057 Composition tests pass.

— posted on behalf of the author via GitHub Copilot CLI

AArnott (approved):

AArnott:
Including the commit I pushed to this PR, here are the new benchmarks

## Main

| Method                  | Mean         | Error       | StdDev      | Gen0     | Gen1     | Allocated  |
|------------------------ |-------------:|------------:|------------:|---------:|---------:|-----------:|
| Discovery               | 2,836.035 μs |  55.0063 μs |  67.5526 μs | 140.6250 |  46.8750 | 2275.93 KB |
| Composition             | 9,660.884 μs | 203.1523 μs | 592.6048 μs | 515.6250 | 343.7500 | 8652.07 KB |
| Serialize               | 2,225.121 μs |  43.1565 μs |  57.6127 μs |  23.4375 |   3.9063 |  386.46 KB |
| Deserialize             |   652.594 μs |  12.2643 μs |  10.2413 μs |  35.1563 |  17.5781 |  582.89 KB |
| Runtime                 | 5,574.956 μs | 110.7282 μs | 265.2977 μs | 343.7500 | 265.6250 | 5681.65 KB |
| GetExportsUnconstrained |     7.191 μs |   0.1438 μs |   0.3002 μs |   1.7395 |   0.0229 |   28.42 KB |
| GetExportsConstrained   |    17.307 μs |   0.3897 μs |   1.1490 μs |   2.5940 |        - |   42.48 KB |

## 6fdcfd2f352 (earlier PR iteration)

| Method                  | Mean            | Error          | StdDev         | Gen0     | Gen1     | Allocated |
|------------------------ |----------------:|---------------:|---------------:|---------:|---------:|----------:|
| Discovery               | 2,760,689.59 ns |  54,700.534 ns |  88,331.173 ns | 140.6250 |  46.8750 | 2330664 B |
| Composition             | 7,290,167.08 ns | 144,943.917 ns | 315,096.170 ns | 328.1250 | 148.4375 | 5527726 B |
| Serialize               | 2,201,074.37 ns |  42,688.711 ns |  67,708.825 ns |  23.4375 |   3.9063 |  395733 B |
| Deserialize             |   649,526.17 ns |  12,735.778 ns |  19,828.073 ns |  35.1563 |  17.5781 |  596879 B |
| Runtime                 | 5,806,609.98 ns | 111,696.442 ns | 307,644.462 ns | 343.7500 | 265.6250 | 5818011 B |
| GetExportsUnconstrained |        73.44 ns |       1.487 ns |       2.935 ns |   0.0014 |        - |      24 B |
| GetExportsConstrained   |     8,708.11 ns |     172.944 ns |     440.198 ns |   0.1831 |        - |    3304 B |

## af47ed6 (avoid array copy completely)

| Method                  | Mean            | Error          | StdDev         | Gen0     | Gen1     | Allocated |
|------------------------ |----------------:|---------------:|---------------:|---------:|---------:|----------:|
| Discovery               | 2,738,969.17 ns |  54,696.838 ns |  78,444.577 ns | 140.6250 |  46.8750 | 2330076 B |
| Composition             | 7,562,219.94 ns | 149,276.150 ns | 280,376.816 ns | 312.5000 | 125.0000 | 5279474 B |
| Serialize               | 2,393,616.07 ns |  47,713.436 ns |  86,037.173 ns |  23.4375 |   7.8125 |  386133 B |
| Deserialize             |   641,542.43 ns |  12,317.500 ns |  13,179.587 ns |  35.1563 |  17.5781 |  596878 B |
| Runtime                 | 5,898,010.97 ns | 117,511.826 ns | 288,258.532 ns | 343.7500 | 265.6250 | 5818011 B |
| GetExportsUnconstrained |        73.29 ns |       1.483 ns |       2.598 ns |   0.0014 |        - |      24 B |
| GetExportsConstrained   |     7,988.07 ns |     156.618 ns |     248.413 ns |        - |        - |      24 B |

src/Microsoft.VisualStudio.Composition/ComposableCatalog.cs
```diff
@@ -239,11 +236,53 @@ from export in openGenericExports
...
+                // No constraints: every matching export qualifies.
+                return exports;
+            }
+
+            int exportIndex = 0;
+            foreach (ExportDefinitionBinding export in exports)
+            {
+                if (!IsExportSatisfiedByAllConstraints(constraints, export.ExportDefinition))
+                {
+                    exports = exports.RemoveAt(exportIndex);
```
olegtk, line 255: Assigning exports inside a loop over exports sounds fragile, it probably works fine now because the list is only shrinking, but I feel like there could be some off by one bug hiding here.
Please add a comment or maybe this can be done slightly differently but still O(n)?

test/Microsoft.VisualStudio.Composition.Benchmarks/CompositionBenchmarks.cs
```diff
@@ -0,0 +1,120 @@
...
+        }
+
+        return count;
+    }
+
+    [Benchmark]
+    public int GetExportsUnconstrained() => this.catalog.GetExports(this.unconstrainedImport).Count;
+
+    [Benchmark]
+    public int GetExportsConstrained() => this.catalog.GetExports(this.constrainedImport).Count;
```
olegtk, line 114: both of these benchmarks don't seem to reject any exports (which is somewhat more expensive path now due to RemoveAt). Would be great to have a benchmark that has a lot of rejections to see the impact?

olegtk (approved):

src/Microsoft.VisualStudio.Composition/ComposableCatalog.cs
AArnott: This isn't a `List<T>`, which I agree would be _very_ problematic to mutate while enumerating. But this is an `ImmutableList<T>`, which is perfectly safe to enumerate and mutate at the same time. 
A clarifying code comment would help here, I agree.

olegtk (approved):

test/Microsoft.VisualStudio.Composition.Benchmarks/CompositionBenchmarks.cs
AArnott: It's true we lack a benchmark for rejected exports, but that's a *very* rare case in general, so it's not that interesting of a benchmark.
If it becomes a more common case, we can add the benchmark and further complexify the code to keep both scenarios fast.

sandyarmstrong (approved):

karpinsn:
Thanks for finishing this up @AArnott!

PR #772, opened 2026-08-21
Optimize export lookup and simple activation

AArnott:
## Summary

- cache exact runtime export lookups and initialized shared root values
- bypass general lifecycle state transitions for non-shared parts with no imports or `OnImportsSatisfied` callbacks
- avoid LINQ and empty-array allocations while resolving constructor arguments through the existing reflection lifecycle
- add BenchmarkDotNet coverage for shared, non-shared, constructor, property, complex, and `ImportMany` activation
- document the investigation, JIT tradeoffs, compiled activation experiment, and staged hybrid-plan ideas in `PERFORMANCE.md`

This branch does not add expression-compiled constructors, setters, or activation plans. That experiment is preserved separately on `perf/compiled-activation-plans` and has no pull request.

## Performance

Matched three-run medians from the IocPerformance scenarios. In the 1-thread variant, one thread performs 500,000 iterations. In the 2-thread variant, two threads share the same export provider and each performs 250,000 iterations concurrently. Each iteration resolves three exports, so both variants perform 1.5 million total resolutions. Values are total wall-clock time; the 2-thread variant measures concurrent throughput and contention rather than individual activation latency. `main` was measured at `26831df5`; the PR column contains only the reduced, non-expression-compilation changes. The MEF2 control measurements differed by no more than 2.5% between runs.

| Scenario | main, 1 thread | PR, 1 thread | Speedup | main, 2 threads | PR, 2 threads | Speedup |
|---|---:|---:|---:|---:|---:|---:|
| Singleton | 7,307 ms | 207 ms | 35.30x | 3,908 ms | 124 ms | 31.52x |
| Transient | 10,367 ms | 1,277 ms | 8.12x | 5,640 ms | 866 ms | 6.51x |
| Combined | 15,656 ms | 5,300 ms | 2.95x | 8,878 ms | 3,544 ms | 2.51x |
| Complex | 40,663 ms | 21,288 ms | 1.91x | 25,576 ms | 15,419 ms | 1.66x |
| Property | 32,091 ms | 21,378 ms | 1.50x | 17,825 ms | 12,535 ms | 1.42x |
| IEnumerable | 28,558 ms | 11,729 ms | 2.43x | 17,237 ms | 8,259 ms | 2.09x |

The largest gains come from cached shared retrieval and the narrow lifecycle bypass for simple non-shared parts. Imported graphs retain 1.4-3.0x improvements from cached root lookup and lower constructor-argument overhead. Expression-compiled constructors, setters, and graph activation are not included in these PR results.

## Validation
- `Microsoft.VisualStudio.Composition.Tests`: 2,064 net8.0 and 2,107 net472 tests passed
- benchmark project builds without warnings for net10.0 and net472
- all six activation benchmarks execute under both target frameworks
- no new expression compilation remains in the pull request diff

## Validation pipeline

[vs-mef validation build 15110883](https://dev.azure.com/devdiv/DevDiv/_build/results?buildId=15110883&view=results)


src/Microsoft.VisualStudio.Composition/RuntimeExportProviderFactory+RuntimeExportProvider.cs
```diff
@@ -72,13 +74,89 @@ select this.CreateExport(
...
+
+                if (lookup.TryGetSharedValue(out value))
+                {
+                    return true;
+                }
+
+                value = this.GetRuntimeExportedValue(lookup, type);
+                lookup.SetSharedValue(value);
+                return true;
+            }
```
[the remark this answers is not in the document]
AArnott: Fixed in `12887c5d`. The fast-path cache is now populated only when the lifecycle tracker has reached `Final`, so a same-thread reentrant self-query can still receive the partial instance without publishing it through the cache. I also added a concurrency regression test that holds `OnImportsSatisfied` open and verifies another request waits for finalization.

PERFORMANCE.md
```diff
@@ -0,0 +1,102 @@
...
+```powershell
+dotnet run --project test\Microsoft.VisualStudio.Composition.Benchmarks\Microsoft.VisualStudio.Composition.Benchmarks.csproj -c Release --framework net10.0 -- --filter "*ActivationBenchmarks*"
+```
+
+Benchmark results should be considered together with startup time, allocation data, and the number of generated or JIT-compiled methods.
+
+## Low-JIT-cost optimizations
+
+The `perf-optimizations` branch and its pull request deliberately avoid adding expression-compiled activation methods. They include:
+
```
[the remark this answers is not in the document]
AArnott: Fixed in `12887c5d`. The durable performance documentation now describes these as the low-JIT-cost optimizations without tying them to the `perf-optimizations` branch name.

src/Microsoft.VisualStudio.Composition/RuntimeExportProviderFactory+RuntimeExportProvider.cs
```diff
@@ -72,13 +74,95 @@ select this.CreateExport(
...
 
+            private RuntimeExportLookup CreateRuntimeExportLookup(Type type, string contractName)
+            {
+                IReadOnlyCollection<RuntimeComposition.RuntimeExport> exports = this.composition.GetExports(contractName);
+                string typeIdentity = ContractNameServices.GetTypeIdentity(type);
+                RuntimeComposition.RuntimeExport? matchingExport = null;
+                int matchingExportCount = 0;
+                foreach (RuntimeComposition.RuntimeExport export in exports)
+                {
+                    if (export.Metadata.TryGetValue(CompositionConstants.ExportTypeIdentityMetadataName, out object? exportedTypeIdentity)
```
olegtk, line 168: **Blocking: preserve `object` cardinality semantics.** The existing lookup deliberately omits the export-type-identity constraint when the requested type is equivalent to `object`. This loop always filters by `System.Object`, so with one object-typed export and another typed export under the same contract, `GetExportedValue<object>(contract)` now silently returns the object export instead of reporting ambiguous cardinality. Please either bypass the fast path for `object` or preserve the unconstrained cardinality check, with a regression test.

src/Microsoft.VisualStudio.Composition/RuntimeExportProviderFactory+RuntimeExportProvider.cs
```diff
@@ -72,13 +74,95 @@ select this.CreateExport(
...
+                }
+
+                PartLifecycleTracker partLifecycle = this.GetOrCreateValue(
+                    lookup.Part!.TypeRef,
+                    lookup.Part.TypeRef,
+                    lookup.Part.SharingBoundary,
+                    EmptyMetadata,
+                    !lookup.Part.IsShared,
+                    nonSharedPartOwner: null);
+                object? part = partLifecycle.GetValueReadyToExpose();
```
olegtk, line 141: **Blocking: preserve exported-member lifecycle behavior.** The established member-export path calls `GetValueReadyToRetrieveExportingMembers()`, while this fast path calls `GetValueReadyToExpose()` before reading a non-static member. That satisfies imports and runs `OnImportsSatisfied`, which can change the exported value or introduce exceptions that did not previously occur. A probe whose exported property observes an import returns `"before"` on the base commit and `"after"` here. Please retain the prior lifecycle behavior for member exports (or exclude them from this fast path) and add coverage.

src/Microsoft.VisualStudio.Composition/RuntimeExportProviderFactory+RuntimeExportProvider.cs
```diff
@@ -478,6 +562,50 @@ internal ValueForImportSite(object? value)
...
+
+                    value = null;
+                    return false;
+                }
+
+                internal void SetSharedValue(object? value)
+                {
+                    if (this.Part!.IsShared && this.Export!.MemberRef is null)
+                    {
+                        this.sharedValue = value;
```
olegtk, line 624: **Blocking: release cached shared values when the provider is disposed.** `RuntimeExportLookup.sharedValue` remains a strong reference after `PartLifecycleTracker.Dispose()` clears the tracker value. Consequently, a disposed provider that remains referenced retains shared parts and their object graphs. A focused weak-reference test collects the shared value on the base commit but leaves it alive on this head. Please clear these cached references during disposal (including when part disposal throws) and add a GC regression test.

olegtk (commented):
I found three observable behavior regressions in the export fast path. Focused regression probes pass on base `8ece0dc8` and fail on this head (in both normal and slow-reflection V3 runs).

src/Microsoft.VisualStudio.Composition/RuntimeExportProviderFactory+RuntimeExportProvider.cs
AArnott: Fixed in f7de9fe4 by bypassing the fast path for requests equivalent to `object`, preserving the original unconstrained cardinality behavior. Added a mixed-contract ambiguity regression test.

src/Microsoft.VisualStudio.Composition/RuntimeExportProviderFactory+RuntimeExportProvider.cs
AArnott: Fixed in f7de9fe4 by excluding exported members from this fast path so they continue through the established member-export lifecycle. Added coverage proving the property is read before imports are satisfied.

src/Microsoft.VisualStudio.Composition/RuntimeExportProviderFactory+RuntimeExportProvider.cs
AArnott: Fixed in f7de9fe4 by clearing the runtime export lookup cache in a `finally` during provider disposal, including when part disposal throws. Added a weak-reference regression test for that exception path.

olegtk (approved):

PR #782, opened 2026-08-31
Reject exports in uncreatable sharing boundaries

### LINKED ISSUE #132: ImportMany failure at runtime instead of composition time
Consider the following scenario having ImportMany over mixed scoped/unscoped exports
```C#
public interface IMessageHandler { }

[Export(typeof(IMessageHandler))]
public class UniqueMessageHandler : IMessageHandler { }

[Export(typeof(IMessageHandler))]
[Shared]
public class GlobalMessageHandler : IMessageHandler { }

[Export(typeof(IMessageHandler))]
[Shared("somewhere")]
public class ScopedMessageHandler : IMessageHandler { }

[Export]
public class Helper1
{
    [ImportMany]
    public ICollection<Lazy<IMessageHandler>> Handlers { get; set; }
}

[Export]
public class Helper2
{
    [Import]
    public Lazy<Helper1> LazyHelper { get; set; }
}
```

Observed behavior:
* in MEF v1 this cannot happen because shared scopes do not exist
* in MEF v2 (the nuget version)
  - `Helper1.Handlers` gets populated with all 3 `IMessageHandler` exports (probably a bug because now application code sees metadata which it should have never seen)
  - fails at runtime when resolving `Lazy<IMessageHandler>.Value` on the scoped import
* in MEF v3 (vs-mef, this project)
  - `Helper1` fails to resolve at runtime because it realizes one of the 3 `IMessageHandlers` is scoped, so it doesn't bother to construct a Lazy import around it. (This is good, even though the failure at runtime is kinda late.)
  - `Helper2.LazyHelper` gets populated with a `Helper1` export even though it will fail to resolve at runtime when calling `Lazy<Helper1>.Value` (This is bad because it delays failure and application code inspecting metadata doesn't know the import is invalid.)

Expected behavior:
* since `Helper1` can never be constructed it should have been pulled from the composition graph and generated a corresponding entry in the `CompositionErrors`

PS: The original scenario was someone marking an export as scoped in an extension assembly, causing random application failures because the ImportMany was lazily resolved. Was kinda hard to root cause.

AArnott:
Named sharing-boundary exports can currently remain in an otherwise valid composition even when no reachable `ExportFactory` can create their scope. Lazy imports then expose exports that fail only when evaluated at runtime, making the underlying composition problem difficult to diagnose.

Detect creatable sharing boundaries from the root composition using a fixed-point traversal. Parts assigned to unreachable boundaries are now reported in `CompositionErrors` and removed by stable-composition salvage, allowing `ImportMany` to retain only usable exports. Regression coverage includes the original lazy `ImportMany` scenario and nested boundaries whose apparent factory belongs to an unreachable parent scope.

Fixes #132

### DIFF
--- a/src/Microsoft.VisualStudio.Composition/CompositionConfiguration.cs
+++ b/src/Microsoft.VisualStudio.Composition/CompositionConfiguration.cs
@@ -140,10 +140,11 @@ namespace Microsoft.VisualStudio.Composition
                 errors.AddRange(part.Validate(metadataViewsAndProviders));
             }
 
-            errors.AddRange(FindPartsInUncreatableSharingBoundaries(parts, sharingBoundaryOverrides));
-
             // Detect loops of all non-shared parts.
             errors.AddRange(FindLoops(parts));
+
+            var initiallyViableParts = parts.Except(errors.SelectMany(error => error.Parts));
+            errors.AddRange(FindPartsInUncreatableSharingBoundaries(initiallyViableParts, sharingBoundaryOverrides));
 
             // If errors are found, re-validate the salvaged parts in case there are parts whose dependencies are affected by the initial errors
             if (errors.Count > 0)
@@ -180,6 +181,9 @@ namespace Microsoft.VisualStudio.Composition
                     {
                         previousErrors.AddRange(part.RemoveSatisfyingExports(invalidPartDefinitionsSet));
                     }
+
+                    var viableParts = salvagedParts.Except(previousErrors.SelectMany(error => error.Parts));
+                    previousErrors.AddRange(FindPartsInUncreatableSharingBoundaries(viableParts, sharingBoundaryOverrides));
                 }
 
                 var finalCatalog = ComposableCatalog.Create(catalog.Resolver).AddParts(salvagedPartDefinitions);
--- a/src/Microsoft.VisualStudio.Composition/CompositionConfiguration.cs
+++ b/src/Microsoft.VisualStudio.Composition/CompositionConfiguration.cs
@@ -462,8 +462,8 @@ namespace Microsoft.VisualStudio.Composition
 
             HashSet<ComposedPart> AnalyzeUnreachableParts(IReadOnlyCollection<ComposedPart> prunedOptionalExports)
             {
-                // The query-directed search normally visits only scopes relevant to one part. Report an
-                // indeterminate result at this bound so pathological graphs cannot exhaust composition-time memory.
+                // The query-directed search normally visits only scopes relevant to one part. Return null
+                // at this bound so the caller does not diagnose an indeterminate result as unreachable.
                 const int MaxScopesPerPart = 1024;
 
                 var requiredSharingBoundaries = partsList.ToDictionary(
--- a/src/Microsoft.VisualStudio.Composition/CompositionConfiguration.cs
+++ b/src/Microsoft.VisualStudio.Composition/CompositionConfiguration.cs
@@ -398,34 +398,117 @@ namespace Microsoft.VisualStudio.Composition
             Requires.NotNull(sharingBoundaryOverrides, nameof(sharingBoundaryOverrides));
 
             var partsList = parts.ToList();
-            var creatableSharingBoundaries = new HashSet<string>();
-            bool addedSharingBoundary;
+            var partsByDefinition = partsList.ToDictionary(part => part.Definition);
+            var requiredSharingBoundaries = partsList.ToDictionary(
+                part => part,
+                part =>
+                {
+                    var boundaries = new HashSet<string>();
+                    string? sharingBoundary = GetEffectiveSharingBoundary(part);
+                    if (!string.IsNullOrEmpty(sharingBoundary))
+                    {
+                        boundaries.Add(sharingBoundary!);
+                    }
+
+                    return boundaries;
+                });
+            var blockedParts = new HashSet<ComposedPart>();
+            bool requirementsChanged;
             do
             {
-                addedSharingBoundary = false;
-                foreach (var part in partsList.Where(part => part.RequiredSharingBoundaries.All(creatableSharingBoundaries.Contains)))
+                requirementsChanged = false;
+                foreach (var part in partsList.Where(part => !blockedParts.Contains(part)))
                 {
-                    var sharingBoundariesCreatedBySatisfiedImports = part.SatisfyingExports
-                        .Where(import => import.Key.IsExportFactory && import.Value.Count > 0)
-                        .SelectMany(import => import.Key.ImportDefinition.ExportFactorySharingBoundaries)
-                        .Where(sharingBoundary => !string.IsNullOrEmpty(sharingBoundary));
-                    foreach (string sharingBoundary in sharingBoundariesCreatedBySatisfiedImports)
+                    foreach (var import in part.SatisfyingExports.Where(import => import.Key.ImportDefinition.Cardinality == ImportCardinality.ExactlyOne))
                     {
-                        addedSharingBoundary |= creatableSharingBoundaries.Add(sharingBoundary);
+                        var viableExports = import.Value
+                            .Where(export => partsByDefinition.TryGetValue(export.PartDefinition, out ComposedPart? exportedPart) && !blockedParts.Contains(exportedPart))
+                            .ToList();
+                        if (viableExports.Count != 1)
+                        {
+                            requirementsChanged |= blockedParts.Add(part);
+                            break;
+                        }
+
+                        if (!import.Key.IsExportFactory || import.Key.ImportDefinition.ExportFactorySharingBoundaries.Count == 0)
+                        {
+                            var exportedPart = partsByDefinition[viableExports[0].PartDefinition];
+                            int previousBoundaryCount = requiredSharingBoundaries[part].Count;
+                            requiredSharingBoundaries[part].UnionWith(requiredSharingBoundaries[exportedPart]);
+                            requirementsChanged |= requiredSharingBoundaries[part].Count != previousBoundaryCount;
+                        }
                     }
                 }
             }
-            while (addedSharingBoundary);
+            while (requirementsChanged);
+
+            var reachableScopes = new List<(ImmutableHashSet<string> AllBoundaries, ImmutableHashSet<string> FreshBoundaries)>
+            {
+                (ImmutableHashSet<string>.Empty, ImmutableHashSet<string>.Empty),
+            };
+            bool scopeAdded;
+            do
+            {
+                scopeAdded = false;
+                foreach (var scope in reachableScopes.ToList())
+                {
+                    foreach (var part in partsList.Where(part => CanInstantiatePartInScope(part, scope)))
+                    {
+                        foreach (var import in part.SatisfyingExports.Where(import => import.Key.IsExportFactory))
+                        {
+                            bool hasViableExport = import.Value.Any(export => partsByDefinition.TryGetValue(export.PartDefinition, out ComposedPart? exportedPart) && !blockedParts.Contains(exportedPart));
+                            var freshBoundaries = import.Key.ImportDefinition.ExportFactorySharingBoundaries
+                                .Where(boundary => !string.IsNullOrEmpty(boundary))
+                                .ToImmutableHashSet();
+                            if (hasViableExport && freshBoundaries.Count > 0)
+                            {
+                                var allBoundaries = scope.AllBoundaries.Union(freshBoundaries);
+                                if (!reachableScopes.Any(existing => existing.AllBoundaries.SetEquals(allBoundaries) && existing.FreshBoundaries.SetEquals(freshBoundaries)))
+                                {
+                                    reachableScopes.Add((allBoundaries, freshBoundaries));
+                                    scopeAdded = true;
+                                }
+                            }
+                        }
+                    }
+                }
+            }
+            while (scopeAdded);
 
             foreach (var part in partsList)
             {
-                string? sharingBoundary = sharingBoundaryOverrides.TryGetValue(part.Definition, out string? effectiveSharingBoundary)
-                    ? effectiveSharingBoundary
-                    : part.Definition.SharingBoundary;
-                if (!string.IsNullOrEmpty(sharingBoundary) && !creatableSharingBoundaries.Contains(sharingBoundary!))
+                string? sharingBoundary = GetEffectiveSharingBoundary(part);
+                if (!string.IsNullOrEmpty(sharingBoundary) && !reachableScopes.Any(scope => CanInstantiatePartInScope(part, scope)))
                 {
                     yield return new ComposedPartDiagnostic(part, Strings.SharingBoundaryHasNoExportFactory, sharingBoundary);
                 }
+            }
+
+            string? GetEffectiveSharingBoundary(ComposedPart part)
+            {
+                return sharingBoundaryOverrides.TryGetValue(part.Definition, out string? effectiveSharingBoundary)
+                    ? effectiveSharingBoundary
+                    : part.Definition.SharingBoundary;
+            }
+
+            bool CanInstantiatePartInScope(
+                ComposedPart part,
+                (ImmutableHashSet<string> AllBoundaries, ImmutableHashSet<string> FreshBoundaries) scope)
+            {
+                if (blockedParts.Contains(part) || !requiredSharingBoundaries[part].IsSubsetOf(scope.AllBoundaries))
+                {
+                    return false;
+                }
+
+                string? sharingBoundary = GetEffectiveSharingBoundary(part);
+                if (!part.Definition.IsShared)
+                {
+                    return true;
+                }
+
+                return string.IsNullOrEmpty(sharingBoundary)
+                    ? scope.AllBoundaries.Count == 0
+                    : scope.FreshBoundaries.Contains(sharingBoundary!);
             }
         }
 
--- a/src/Microsoft.VisualStudio.Composition/CompositionConfiguration.cs
+++ b/src/Microsoft.VisualStudio.Composition/CompositionConfiguration.cs
@@ -406,6 +406,55 @@ namespace Microsoft.VisualStudio.Composition
             var partsByDefinition = partsList.ToDictionary(part => part.Definition, ReferenceEquality<ComposablePartDefinition>.Default);
             var pessimisticUnreachableParts = AnalyzeUnreachableParts(ImmutableHashSet<ComposedPart>.Empty);
             var unreachableParts = AnalyzeUnreachableParts(pessimisticUnreachableParts);
+            while (true)
+            {
+                var currentlyUnreachableParts = AnalyzeUnreachableParts(unreachableParts);
+                currentlyUnreachableParts.ExceptWith(unreachableParts);
+                if (currentlyUnreachableParts.Count == 0)
+                {
+                    break;
+                }
+
+                // Reject terminal groups first. Parts that merely depend on a rejected optional export
+                // are reconsidered after that export is removed from the graph.
+                var optionalDependencies = currentlyUnreachableParts.ToDictionary(
+                    part => part,
+                    part => new HashSet<ComposedPart>(part.SatisfyingExports
+                        .Where(import => import.Key.ImportDefinition.Cardinality != ImportCardinality.ExactlyOne
+                            && (!import.Key.IsExportFactory || import.Key.ImportDefinition.ExportFactorySharingBoundaries.Count == 0))
+                        .SelectMany(import => import.Value)
+                        .Select(export => partsByDefinition.TryGetValue(export.PartDefinition, out ComposedPart? exportedPart) ? exportedPart : null)
+                        .OfType<ComposedPart>()
+                        .Where(currentlyUnreachableParts.Contains)));
+                var reachableDependenciesByPart = currentlyUnreachableParts.ToDictionary(part => part, GetReachableDependencies);
+                foreach (var part in currentlyUnreachableParts)
+                {
+                    if (reachableDependenciesByPart[part].All(dependency => reachableDependenciesByPart[dependency].Contains(part)))
+                    {
+                        unreachableParts.Add(part);
+                    }
+                }
+
+                HashSet<ComposedPart> GetReachableDependencies(ComposedPart part)
+                {
+                    var reachableDependencies = new HashSet<ComposedPart> { part };
+                    var pendingDependencies = new Stack<ComposedPart>();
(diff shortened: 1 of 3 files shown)


src/Microsoft.VisualStudio.Composition/CompositionConfiguration.cs
```diff
@@ -384,6 +386,36 @@ from export in importAndExports.Value
...
+            do
+            {
+                addedSharingBoundary = false;
+                foreach (var part in partsList.Where(part => part.RequiredSharingBoundaries.All(creatableSharingBoundaries.Contains)))
+                {
+                    foreach (string sharingBoundary in part.Definition.Imports.SelectMany(import => import.ImportDefinition.ExportFactorySharingBoundaries))
+                    {
+                        addedSharingBoundary |= creatableSharingBoundaries.Add(sharingBoundary);
+                    }
+                }
```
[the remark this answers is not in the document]
AArnott: Updated boundary reachability to consider only satisfied `ExportFactory` imports and to ignore empty boundary names. Added a regression where an optional factory has no matching export, confirming that its declared scope remains uncreatable and the scoped part is rejected.

> [!NOTE]
> Auto-replied by the [GitHub Copilot app](https://github.com/github/app).

src/Microsoft.VisualStudio.Composition/CompositionConfiguration.cs
```diff
@@ -384,6 +390,45 @@ from export in importAndExports.Value
...
+            Requires.NotNull(parts, nameof(parts));
+            Requires.NotNull(sharingBoundaryOverrides, nameof(sharingBoundaryOverrides));
+
+            var partsList = parts.ToList();
+            var creatableSharingBoundaries = new HashSet<string>();
+            bool addedSharingBoundary;
+            do
+            {
+                addedSharingBoundary = false;
+                foreach (var part in partsList.Where(part => part.RequiredSharingBoundaries.All(creatableSharingBoundaries.Contains)))
```
olegtk: **Blocking: recompute boundary requirements after salvage pruning.** `RequiredSharingBoundaries` was computed from the original satisfying-export graph and remains stale after `RemoveSatisfyingExports` prunes optional or `ImportMany` exports. Repro: a root part imports both a global service and an uncreatable scoped service via `ImportMany`, and also owns an `ExportFactory` for an otherwise valid scope. Because the stale requirement makes the root appear unreachable, this pass marks the valid scoped part uncreatable; salvage then invalidates the required factory import and removes the root as well. The root and valid scope should survive once the unusable `ImportMany` export is pruned. Please derive requirements from the current satisfying exports, including during initial diagnosis, or recompute the dependency closure before each reachability pass, with regression coverage for a surviving `ImportMany` consumer that owns another boundary factory.

src/Microsoft.VisualStudio.Composition/CompositionConfiguration.cs
```diff
@@ -384,6 +390,45 @@ from export in importAndExports.Value
...
 
+        private static IEnumerable<ComposedPartDiagnostic> FindPartsInUncreatableSharingBoundaries(
+            IEnumerable<ComposedPart> parts,
+            ImmutableDictionary<ComposablePartDefinition, string> sharingBoundaryOverrides)
+        {
+            Requires.NotNull(parts, nameof(parts));
+            Requires.NotNull(sharingBoundaryOverrides, nameof(sharingBoundaryOverrides));
+
+            var partsList = parts.ToList();
+            var creatableSharingBoundaries = new HashSet<string>();
```
olegtk: **Blocking: a global union of boundary names loses reachable scope topology.** Repro: a root has separate factories creating sibling scopes A and B; another part requires both A and B and owns a factory for C. This set eventually contains both A and B, so the A+B part is treated as reachable and C is marked creatable, although no reachable provider scope contains A and B together. The C-scoped export consequently remains in the composition even though its only apparent factory cannot be reached. Please track reachable boundary combinations or parent paths rather than a single union of names, and add a sibling-scopes test distinct from a factory that explicitly creates both boundaries.

olegtk (commented):
I found two blocking reachability regressions. Both reproduce with focused composition tests on this head; the existing suite does not cover them.

src/Microsoft.VisualStudio.Composition/CompositionConfiguration.cs
AArnott: Reworked reachability to derive required boundaries from the current satisfying-export graph and to rerun after every salvage pass. Optional and `ImportMany` edges no longer over-constrain a surviving factory owner; the requested regression confirms its valid scope and root part remain usable after the bad export is pruned.

> [!NOTE]
> Auto-replied by the [GitHub Copilot app](https://github.com/github/app).

src/Microsoft.VisualStudio.Composition/CompositionConfiguration.cs
AArnott: Replaced the global boundary-name union with reachable provider scope nodes that preserve inherited boundaries and each factory's jointly-created fresh boundaries. Added the sibling A/B repro, which now rejects C while preserving both valid sibling scopes.

> [!NOTE]
> Auto-replied by the [GitHub Copilot app](https://github.com/github/app).

src/Microsoft.VisualStudio.Composition/CompositionConfiguration.cs
```diff
@@ -384,6 +390,128 @@ from export in importAndExports.Value
...
 
+        private static IEnumerable<ComposedPartDiagnostic> FindPartsInUncreatableSharingBoundaries(
+            IEnumerable<ComposedPart> parts,
+            ImmutableDictionary<ComposablePartDefinition, string> sharingBoundaryOverrides)
+        {
+            Requires.NotNull(parts, nameof(parts));
+            Requires.NotNull(sharingBoundaryOverrides, nameof(sharingBoundaryOverrides));
+
+            var partsList = parts.ToList();
+            var partsByDefinition = partsList.ToDictionary(part => part.Definition);
```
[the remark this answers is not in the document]
AArnott: Updated the lookup to use `ReferenceEquality<ComposablePartDefinition>.Default`, matching the surrounding graph construction and avoiding structural equality costs for large catalogs.

> [!NOTE]
> Auto-replied by the [GitHub Copilot app](https://github.com/github/app).

src/Microsoft.VisualStudio.Composition/CompositionConfiguration.cs
```diff
@@ -384,6 +390,128 @@ from export in importAndExports.Value
...
+                    return boundaries;
+                });
+            var blockedParts = new HashSet<ComposedPart>();
+            bool requirementsChanged;
+            do
+            {
+                requirementsChanged = false;
+                foreach (var part in partsList.Where(part => !blockedParts.Contains(part)))
+                {
+                    foreach (var import in part.SatisfyingExports.Where(import => import.Key.ImportDefinition.Cardinality == ImportCardinality.ExactlyOne))
```
olegtk: **Blocking: surviving optional and `ImportMany` exports must contribute scope requirements.** This traversal propagates requirements only through `ExactlyOne` imports, but selected optional and `ImportMany` exports are still resolved by the runtime. Repro: root factories create sibling scopes A and B; a non-shared part imports A- and B-scoped services via `ImportMany` and owns a factory for Nested. No provider scope contains both A and B, so that owner cannot be instantiated, yet it is treated as root-instantiable and Nested remains in the composition. Please account for the requirements of all surviving selected exports while retaining the intended pruning behavior, and add coverage for `ImportMany` spanning incompatible sibling scopes.

src/Microsoft.VisualStudio.Composition/CompositionConfiguration.cs
```diff
@@ -384,6 +390,128 @@ from export in importAndExports.Value
...
+            do
+            {
+                scopeAdded = false;
+                foreach (var scope in reachableScopes.ToList())
+                {
+                    foreach (var part in partsList.Where(part => CanInstantiatePartInScope(part, scope)))
+                    {
+                        foreach (var import in part.SatisfyingExports.Where(import => import.Key.IsExportFactory))
+                        {
+                            bool hasViableExport = import.Value.Any(export => partsByDefinition.TryGetValue(export.PartDefinition, out ComposedPart? exportedPart) && !blockedParts.Contains(exportedPart));
```
olegtk: **Blocking: validate each factory target in the prospective child scope.** `hasViableExport` only checks that a target part exists and is not globally blocked. Repro: a root factory creates scope C; a separate root factory creates scope A but targets a non-shared part that requires C. The target is viable in C, but invoking the A factory from root creates only A and must fail because C is absent. This code nevertheless marks A reachable and preserves unrelated A-scoped exports. Please require at least one target export to be instantiable in `scope.AllBoundaries ∪ freshBoundaries`, and add this sibling-scope target regression.

src/Microsoft.VisualStudio.Composition/CompositionConfiguration.cs
```diff
@@ -384,6 +390,128 @@ from export in importAndExports.Value
...
+                            var freshBoundaries = import.Key.ImportDefinition.ExportFactorySharingBoundaries
+                                .Where(boundary => !string.IsNullOrEmpty(boundary))
+                                .ToImmutableHashSet();
+                            if (hasViableExport && freshBoundaries.Count > 0)
+                            {
+                                var allBoundaries = scope.AllBoundaries.Union(freshBoundaries);
+                                if (!reachableScopes.Any(existing => existing.AllBoundaries.SetEquals(allBoundaries) && existing.FreshBoundaries.SetEquals(freshBoundaries)))
```
olegtk: **Blocking: reachable-scope enumeration has exponential growth.** A non-shared, boundary-free factory owner is eligible in every generated scope. With `k` independent single-boundary factories, this materializes `1 + k * 2^(k-1)` `(AllBoundaries, FreshBoundaries)` states; 20 boundaries exceed 10 million states. Each insertion also linearly scans the list with `Any`, making the total work substantially worse. This is a composition-time hang or memory-exhaustion risk for large catalogs. Please use graph-directed or symbolic reachability that avoids enumerating the boundary power set, with a scale regression test or explicit complexity bound.

olegtk (commented):
The two prior scenarios are fixed, but the replacement scope model introduces three additional blocking issues. The first two reproduce with focused composition tests on this head; the third follows directly from the generated state space.

src/Microsoft.VisualStudio.Composition/CompositionConfiguration.cs
AArnott: Implemented a two-pass analysis: the first pass identifies pessimistically unreachable optional/`ImportMany` candidates, and the second prunes those candidates while retaining requirements from all surviving selected exports. Added the incompatible sibling-scope `ImportMany` repro; its nested scope is now rejected.

> [!NOTE]
> Auto-replied by the [GitHub Copilot app](https://github.com/github/app).

src/Microsoft.VisualStudio.Composition/CompositionConfiguration.cs
AArnott: Factory edges now require a target export to be instantiable in the prospective child scope before that scope is considered reachable. Added the sibling target regression where an A-factory target requires C; unrelated A-scoped exports are now rejected while C remains usable.

> [!NOTE]
> Auto-replied by the [GitHub Copilot app](https://github.com/github/app).

src/Microsoft.VisualStudio.Composition/CompositionConfiguration.cs
AArnott: Replaced eager global scope enumeration with query-directed searches restricted to boundaries relevant to each candidate part. Each search has an explicit 1,024-scope bound and fails open on overflow, bounding memory and preventing false composition errors on pathological graphs.

> [!NOTE]
> Auto-replied by the [GitHub Copilot app](https://github.com/github/app).

test/Microsoft.VisualStudio.Composition.Tests/UncreatableSharingBoundaryTests.cs
```diff
@@ -0,0 +1,377 @@
...
+
+namespace Microsoft.VisualStudio.Composition.Tests;
+
+using System;
+using System.Collections.Generic;
+using System.Composition;
+using System.Linq;
+using Xunit;
+using MefV1 = System.ComponentModel.Composition;
+
```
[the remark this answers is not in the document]
AArnott: No change is needed here: C# permits `using` directives after a file-scoped namespace, where they are scoped within that namespace. This exact file compiled successfully again on the current head, and all 20 `UncreatableSharingBoundaryTests` passed.

> [!NOTE]
> Auto-replied by the [GitHub Copilot app](https://github.com/github/app).

olegtk (commented):
The previous three scenarios are addressed, but the bounded two-pass replacement has two additional blocking correctness issues. Both reproduce with focused tests on the current head.

src/Microsoft.VisualStudio.Composition/CompositionConfiguration.cs
```diff
@@ -384,6 +390,203 @@ from export in importAndExports.Value
...
+
+            var partsList = parts.ToList();
+            if (!partsList.Any(part => !string.IsNullOrEmpty(GetEffectiveSharingBoundary(part))))
+            {
+                yield break;
+            }
+
+            var partsByDefinition = partsList.ToDictionary(part => part.Definition, ReferenceEquality<ComposablePartDefinition>.Default);
+            var pessimisticUnreachableParts = AnalyzeUnreachableParts(ImmutableHashSet<ComposedPart>.Empty);
+            var unreachableParts = AnalyzeUnreachableParts(pessimisticUnreachableParts);
```
olegtk, line 408: **Blocking: the second pass can retain exports while still analyzing their optional edges as pruned.** The second call excludes optional and `ImportMany` exports targeting every first-pass-unreachable part when computing requirements. If removing those edges makes the parts reachable, they disappear from the final unreachable set, but their actual `SatisfyingExports` are never changed. Repro: root factories create sibling scopes A and B; the A-scoped and B-scoped parts optionally import each other. Pass one marks both unreachable; pass two suppresses both cross-edges and declares both reachable, so the configuration has no errors. At runtime each retained optional import is still materialized and factory activation fails because the sibling boundary is absent. Please compute a self-consistent retained/pruned graph and either persist the pruned import edges or reject the affected parts, with this optional-cycle regression.

src/Microsoft.VisualStudio.Composition/CompositionConfiguration.cs
```diff
@@ -384,6 +390,203 @@ from export in importAndExports.Value
...
+                                var allBoundaries = scope.AllBoundaries.Union(freshBoundaries);
+                                var childScope = (AllBoundaries: allBoundaries, FreshBoundaries: freshBoundaries);
+                                bool hasInstantiableTarget = import.Value
+                                    .Where(export => partsByDefinition.TryGetValue(export.PartDefinition, out ComposedPart? exportedPart) && !blockedParts.Contains(exportedPart))
+                                    .Any(export => CanInstantiatePartInScope(partsByDefinition[export.PartDefinition], childScope));
+                                if (hasInstantiableTarget && !reachableScopes.Any(existing => existing.AllBoundaries.SetEquals(allBoundaries) && existing.FreshBoundaries.SetEquals(freshBoundaries)))
+                                {
+                                    if (reachableScopes.Count >= MaxScopesPerPart)
+                                    {
+                                        return true;
```
olegtk: **Blocking: scope-budget exhaustion is not proof of reachability.** Returning `true` at the 1,024-state limit silently preserves an unreachable part. Repro: ten independently nestable boundary factories plus a part that requires all ten and belongs to an additional boundary with no factory. The search exceeds the cap in about four seconds and retains that part even though its own boundary is provably unreachable. This defeats the composition-time guarantee for a catalog size that is not pathological. Please represent exhaustion as indeterminate or use a sound bounded analysis; do not convert it to reachable. Add a scale regression that confirms an unreachable boundary remains rejected when the search budget is reached.

AArnott: Fixed in `22d0b4e6`. Budget exhaustion is now represented as indeterminate rather than reachable, and the solver first soundly rejects a target whose boundary has no viable satisfied factory, avoiding the exponential search entirely for this case. Added scale regressions for both a missing factory and a nominal factory whose only target is blocked.

> [!NOTE]
> Auto-replied by the [GitHub Copilot app](https://github.com/github/app).

src/Microsoft.VisualStudio.Composition/CompositionConfiguration.cs
AArnott: Fixed in `22d0b4e6`. The analysis now starts with the two-pass definite rejection set, then repeatedly rejects terminal optional-dependency groups and reanalyzes dependents, so mutually rescuing cycles are removed while acyclic importers can still be salvaged after their rejected exports are pruned. Added the sibling-scope optional-cycle regression.

> [!NOTE]
> Auto-replied by the [GitHub Copilot app](https://github.com/github/app).

src/Microsoft.VisualStudio.Composition/CompositionConfiguration.cs
```diff
@@ -384,6 +390,260 @@ from export in importAndExports.Value
...
+                    };
+                    for (int scopeIndex = 0; scopeIndex < reachableScopes.Count; scopeIndex++)
+                    {
+                        var scope = reachableScopes[scopeIndex];
+                        if (CanInstantiatePartInScope(targetPart, scope))
+                        {
+                            return true;
+                        }
+
+                        foreach (var factoryOwner in partsList.Where(part => CanInstantiatePartInScope(part, scope)))
```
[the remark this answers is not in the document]
AArnott: The current behavior is intentional. An `ExportFactory` captures the provider where its importing owner was instantiated; invoking an ancestor-owned factory while in a descendant scope still creates from that ancestor provider, so it cannot extend the descendant scope. The search therefore considers that factory at its owner’s scope, where the corresponding runtime transition originates.

> [!NOTE]
> Auto-replied by the [GitHub Copilot app](https://github.com/github/app).

olegtk (commented):
Two blocking correctness issues remain on the current head. Details and reproductions are inline; I recommend resolving both before approval.

src/Microsoft.VisualStudio.Composition/CompositionConfiguration.cs
```diff
@@ -384,6 +390,260 @@ from export in importAndExports.Value
...
+                            {
+                                requirementsChanged |= blockedParts.Add(part);
+                                break;
+                            }
+
+                            foreach (var export in viableExports)
+                            {
+                                var exportedPart = partsByDefinition[export.PartDefinition];
+                                int previousBoundaryCount = requiredSharingBoundaries[part].Count;
+                                requiredSharingBoundaries[part].UnionWith(requiredSharingBoundaries[exportedPart]);
```
olegtk: **Blocking: compute blocked parts before propagating boundary requirements.** This loop grows `requiredSharingBoundaries` while `blockedParts` is still changing. If an optional export contributes a boundary and is later blocked through a required dependency, its contribution is never removed because these sets only grow. The result depends on part/hash-set iteration order.

I reproduced this with an invalid `Bad <- Z <- Y` chain, where `Y` also requires boundary A, a B-scoped `X` imports many `Y`, and the root owns valid A and B factories. Across 20 calls with the same catalog, 11 correctly retained `X` with an empty collection while 9 incorrectly removed `X`, the root, and the valid A-scoped service.

Please compute the blocked-part fixed point first, then rebuild the boundary requirements from only unblocked exports. An order-independence regression should cover this salvage case.

src/Microsoft.VisualStudio.Composition/CompositionConfiguration.cs
```diff
@@ -384,6 +390,260 @@ from export in importAndExports.Value
...
+                                var allBoundaries = scope.AllBoundaries.Union(freshBoundaries);
+                                var childScope = (AllBoundaries: allBoundaries, FreshBoundaries: freshBoundaries);
+                                bool hasInstantiableTarget = import.Value
+                                    .Where(export => partsByDefinition.TryGetValue(export.PartDefinition, out ComposedPart? exportedPart) && !blockedParts.Contains(exportedPart))
+                                    .Any(export => CanInstantiatePartInScope(partsByDefinition[export.PartDefinition], childScope));
+                                if (hasInstantiableTarget && !reachableScopes.Any(existing => existing.AllBoundaries.SetEquals(allBoundaries) && existing.FreshBoundaries.SetEquals(freshBoundaries)))
+                                {
+                                    if (reachableScopes.Count >= MaxScopesPerPart)
+                                    {
+                                        return null;
```
olegtk: **Blocking: search-budget exhaustion is still treated as success.** Returning `null` here is indeterminate, but the caller only rejects `IsPartReachable(part) == false`; consequently an impossible part is silently retained when this limit is reached.

I reproduced this with a 13-part catalog containing ten independently nestable boundaries and an impossible X-scoped part whose boundary has a nominal factory. The search exhausted the limit in about 1.9 seconds, composition reported no errors, and invoking the retained factory then threw `CompositionFailedException` because an import could not be satisfied.

Please use a symbolic/dominance-based search that avoids enumerating the boundary power set, or treat an indeterminate result as a non-salvageable composition failure rather than valid reachability. Add a regression where the target boundary has a factory but the search reaches its budget.

src/Microsoft.VisualStudio.Composition/CompositionConfiguration.cs
AArnott: Fixed in `fc13af84`. Invalid parts now reach a fixed point before any sharing-boundary requirements are propagated, so boundaries from subsequently blocked optional exports cannot remain in monotonic requirement sets. Added a sibling-scope salvage regression whose shared root makes the old iteration-order-dependent implementation reject the otherwise valid B-scoped importer.

> [!NOTE]
> Auto-replied by the [GitHub Copilot app](https://github.com/github/app).

src/Microsoft.VisualStudio.Composition/CompositionConfiguration.cs
AArnott: Fixed in `fc13af84`. Scope search now explores the most capable states first and prunes states dominated by another state with the same freshly owned boundaries and a superset of available boundaries; the bounded fallback rejects rather than preserving an indeterminate runtime-invalid export. Added scaled coverage where a nominal factory targets an impossible part and is removed from the salvaged graph.

> [!NOTE]
> Auto-replied by the [GitHub Copilot app](https://github.com/github/app).

olegtk (approved):

PR #794, opened 2026-09-11
Composition: reduce sharing boundary analysis overhead

AArnott:
## Summary

- stop uncreatable-sharing-boundary validation after the first reachability pass when a valid catalog has no unreachable parts
- replace allocation-heavy LINQ pipelines in this analysis with direct loops
- enumerate the known `ImmutableDictionary`/`ImmutableList` representations without interface-enumerator boxing, with an indexed fallback for other `IReadOnlyList` implementations
- avoid materializing viable export lists and stop required-import cardinality checks after a second viable export is found

## Why

The VS `Startup.BuildingMEFCache` Speedometer trace for the vs-mef 18.9.74 insertion attributes at least 402.5 MB (91.9% of the matched trace pair's net allocation increase) to `FindPartsInUncreatableSharingBoundaries`. Valid catalogs ran the reachability analysis three times even when the first pass found nothing to prune.

PIT: https://dev.azure.com/devdiv/DevDiv/_apps/hub/ms-vseng.pit-vsengPerf.pit-hub?targetBuild=12211.112.Build_0bdbc590.260911.170238.779226&targetBranch=main&targetPerfBuildId=15295393&runGroup=Speedometer&baselineBuild=12211.55&baselineBranch=main&selectedCounters=Startup.BuildingMEFCache%3E9990.Totals%3ECLR_BytesAllocated_devenv

## Performance

BenchmarkDotNet `CompositionBenchmarks.Composition`, measured sequentially on the same idle machine against merged commit `be17f200`:

| | Mean | Allocated |
|---|---:|---:|
| Baseline | 30.60 ms | 7.17 MB |
| This change | 13.54 ms | 5.22 MB |
| Improvement | 55.7% | 27.2% |

For context, a validation-disabled control measured 13.15 ms / 5.17 MB, so this retains the validation at nearly the control's common-case cost.

## Validation

- `Microsoft.VisualStudio.Composition.Tests` net8.0: 2,173 passed
- `Microsoft.VisualStudio.Composition.Tests` net472: 2,217 passed
- builds completed with zero warnings and zero errors

### DIFF
--- a/src/Microsoft.VisualStudio.Composition/CompositionConfiguration.cs
+++ b/src/Microsoft.VisualStudio.Composition/CompositionConfiguration.cs
@@ -424,14 +424,34 @@ namespace Microsoft.VisualStudio.Composition
             Requires.NotNull(parts, nameof(parts));
             Requires.NotNull(sharingBoundaryOverrides, nameof(sharingBoundaryOverrides));
 
-            var partsList = parts.ToList();
-            if (!partsList.Any(part => !string.IsNullOrEmpty(GetEffectiveSharingBoundary(part))))
+            var partsList = new List<ComposedPart>(parts);
+            bool hasSharingBoundary = false;
+            foreach (ComposedPart part in partsList)
+            {
+                if (!string.IsNullOrEmpty(GetEffectiveSharingBoundary(part)))
+                {
+                    hasSharingBoundary = true;
+                    break;
+                }
+            }
+
+            if (!hasSharingBoundary)
             {
                 yield break;
             }
 
-            var partsByDefinition = partsList.ToDictionary(part => part.Definition, ReferenceEquality<ComposablePartDefinition>.Default);
+            var partsByDefinition = new Dictionary<ComposablePartDefinition, ComposedPart>(partsList.Count, ReferenceEquality<ComposablePartDefinition>.Default);
+            foreach (ComposedPart part in partsList)
+            {
+                partsByDefinition.Add(part.Definition, part);
+            }
+
             var pessimisticUnreachableParts = AnalyzeUnreachableParts(ImmutableHashSet<ComposedPart>.Empty);
+            if (pessimisticUnreachableParts.Count == 0)
+            {
+                yield break;
+            }
+
             var unreachableParts = AnalyzeUnreachableParts(pessimisticUnreachableParts);
             while (true)
             {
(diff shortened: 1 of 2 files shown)


src/Microsoft.VisualStudio.Composition/CompositionConfiguration.cs
```diff
@@ -424,14 +424,34 @@ private static IEnumerable<ComposedPartDiagnostic> FindPartsInUncreatableSharing
             Requires.NotNull(parts, nameof(parts));
             Requires.NotNull(sharingBoundaryOverrides, nameof(sharingBoundaryOverrides));
 
-            var partsList = parts.ToList();
-            if (!partsList.Any(part => !string.IsNullOrEmpty(GetEffectiveSharingBoundary(part))))
+            var partsList = new List<ComposedPart>(parts);
```
davkean, line 428: Nit: Perfect to be extracted into a method, the assignment to hasSharingBoundary in the foreach + break, would be much cleaner in a function that returns true/false.

src/Microsoft.VisualStudio.Composition/CompositionConfiguration.cs
```diff
@@ -444,19 +464,41 @@ private static IEnumerable<ComposedPartDiagnostic> FindPartsInUncreatableSharing
...
+                foreach (ComposedPart part in currentlyUnreachableParts)
+                {
+                    reachableDependenciesByPart.Add(part, GetReachableDependencies(part));
+                }
+
                 foreach (var part in currentlyUnreachableParts)
                 {
-                    if (reachableDependenciesByPart[part].All(dependency => reachableDependenciesByPart[dependency].Contains(part)))
+                    bool isTerminalGroup = true;
+                    foreach (ComposedPart dependency in reachableDependenciesByPart[part])
```
davkean: Nit: Another perfect for extraction into a method.

src/Microsoft.VisualStudio.Composition/CompositionConfiguration.cs
```diff
@@ -772,6 +816,204 @@ bool CanInstantiatePartInScope(
...
+        }
+
+        private static bool HasExactlyOneViableExport(
+            IReadOnlyList<ExportDefinitionBinding> exports,
+            Dictionary<ComposablePartDefinition, ComposedPart> partsByDefinition,
+            HashSet<ComposedPart> blockedParts,
+            IReadOnlyCollection<ComposedPart> prunedOptionalExports)
+        {
+            int viableExportCount = 0;
+            if (exports is ImmutableList<ExportDefinitionBinding> immutableExports)
```
davkean: Can you change the API to return an ImmutableList to avoid this awkward if/else? If this is not the only case that would benefit, I would consider introducing a struct enumerable that simplified handling this pattern, similar to internal's PerformanceUtilities' GetStructEnumerable.

src/Microsoft.VisualStudio.Composition/CompositionConfiguration.cs
```diff
@@ -772,6 +816,204 @@ bool CanInstantiatePartInScope(
...
+                }
+            }
+
+            return changed;
+
+            void AddExportBoundaries(ExportDefinitionBinding export)
+            {
+                if (IsViableExport(export, cardinality, partsByDefinition, blockedParts, prunedOptionalExports, out ComposedPart? exportedPart))
+                {
+                    changed |= AddAll(boundaries, requiredSharingBoundaries[exportedPart!]);
```
davkean: IsViableExport is missing annotation that `exportedPart` is non-null when it returns `true`.

davkean (approved):

src/Microsoft.VisualStudio.Composition/CompositionConfiguration.cs
AArnott: Changed the internal map to use `ImmutableList<ExportDefinitionBinding>` values, so these hot-path helpers now enumerate the concrete type directly with no downcast or fallback branch. The public `SatisfyingExports` signature remains unchanged to avoid a binary-breaking API change, and no struct enumerable wrapper was introduced.

<sub>Auto-replied by the [GitHub Copilot app](https://gh.io/app)</sub>

src/Microsoft.VisualStudio.Composition/CompositionConfiguration.cs
AArnott: Added `[NotNullWhen(true)]` to the `exportedPart` out parameter and removed the null-forgiving operator at the call site.

<sub>Auto-replied by the [GitHub Copilot app](https://gh.io/app)</sub>

src/Microsoft.VisualStudio.Composition/CompositionConfiguration.cs
AArnott: Extracted the reciprocal-dependency check into `IsTerminalGroup`, leaving the caller focused on adding terminal groups to `unreachableParts`.

<sub>Auto-replied by the [GitHub Copilot app](https://gh.io/app)</sub>

src/Microsoft.VisualStudio.Composition/CompositionConfiguration.cs
AArnott: Extracted this into `HasSharingBoundary`, which now returns immediately when it finds an effective sharing boundary.

<sub>Auto-replied by the [GitHub Copilot app](https://gh.io/app)</sub>

src/Microsoft.VisualStudio.Composition/CompositionConfiguration.cs
AArnott: Copilot's fix was no good because it required duplicating an entire ImmutableDictionary (that contains the ImmutableList). Downcasting is a very small price to pay to avoid duplicate collections in every ComposedPart.

davkean (approved):

PR #795, opened 2026-09-17
Reduce allocations during attributed part discovery

davkean:
Attributed discovery repeats reflection while building parts. It materializes filtered export-attribute arrays for non-exporting members, reads export metadata twice to identify defining assemblies, and retrieves MEF v2 import attributes twice per property. Avoid those redundant paths while preserving derived export attributes in both compatibility modes.

The existing `Discovery` benchmark measured about 12.3% less allocation on .NET Framework and .NET 10. Median elapsed time per operation improved by 12.8% and 4.5%, respectively. CPU profiles confirm that the separate metadata-assembly reflection pass was removed; they were not used for a raw CPU-sample comparison because the diagnostic runs used separately calibrated invocation counts.

| Runtime | Elapsed time, mean +/- SD (median) | Allocated/op | Gen0/1,000 ops | Gen1/1,000 ops | Gen2/1,000 ops |
| --- | ---: | ---: | ---: | ---: | ---: |
| .NET Framework 4.8.1 base | 5.6836 +/- 0.8709 ms (5.5886 ms) | 4,131,683 B | 654.2969 | 298.1771 | 0 |
| .NET Framework 4.8.1 candidate | 4.8580 +/- 0.5666 ms (4.8732 ms) | 3,623,194 B | 572.9167 | 164.7135 | 0 |
| .NET 10 base | 1.8907 +/- 0.2046 ms (1.8352 ms) | 2,473,910 B | 131.5104 | 55.3385 | 0 |
| .NET 10 candidate | 1.7566 +/- 0.0740 ms (1.7518 ms) | 2,167,979 B | 113.2812 | 44.2708 | 0 |


davkean:
Copilot: Did you microbench this? Please disclose the results. I would like GC/CPU both measured.

src/Microsoft.VisualStudio.Composition/Configuration/AttributedPartDiscovery.cs
```diff
@@ -64,6 +64,7 @@ protected BindingFlags PublicVsNonPublicFlags
             BindingFlags instanceLocal = BindingFlags.DeclaredOnly | BindingFlags.Instance | this.PublicVsNonPublicFlags;
             var declaredProperties = partTypeInfo.GetProperties(instanceLocal);
             var exportingProperties = from member in declaredProperties
+                                      where member.IsAttributeDefined<ExportAttribute>()
```
davkean, line 70: Does imports/part and metadata have the same problem?

davkean:
Benchmarked the exact PR base (`c3ed20db5552f9a3cf500900d7f91425d40567b7`) against the updated candidate (`810ebd78f0ba5c5476e0ff25222c4ddf8b43a2bb`) using the existing `CompositionBenchmarks.Discovery` benchmark in Release. Each framework had two independent runs with 3 launches per run, 5 warmups, and 15 measured iterations per launch.

| Runtime | Elapsed time, mean +/- SD (median) | Allocated/op | Gen0/1,000 ops | Gen1/1,000 ops | Gen2/1,000 ops |
| --- | ---: | ---: | ---: | ---: | ---: |
| .NET Framework 4.8.1 base | 5.6836 +/- 0.8709 ms (5.5886 ms) | 4,131,683 B | 654.2969 | 298.1771 | 0 |
| .NET Framework 4.8.1 candidate | 4.8580 +/- 0.5666 ms (4.8732 ms) | 3,623,194 B | 572.9167 | 164.7135 | 0 |
| .NET 10 base | 1.8907 +/- 0.2046 ms (1.8352 ms) | 2,473,910 B | 131.5104 | 55.3385 | 0 |
| .NET 10 candidate | 1.7566 +/- 0.0740 ms (1.7518 ms) | 2,167,979 B | 113.2812 | 44.2708 | 0 |

The candidate reduces allocation by 12.31% on .NET Framework and 12.37% on .NET 10. Median elapsed time per operation improves by 12.80% and 4.54%, respectively. BenchmarkDotNet retained 90 measured iteration results per .NET Framework variant and 85 base / 86 candidate results on .NET 10 after outlier processing.

`CPUUsageDiagnoser` traces were also collected. They confirm that the separate `GetAssemblyNamesFromMetadataAttributes` path was removed after assembly names were collected during the existing metadata pass. Raw CPU sample totals were not compared because BenchmarkDotNet calibrated invocation counts separately for the diagnostic runs; the table reports elapsed time per operation, not normalized CPU time.

src/Microsoft.VisualStudio.Composition/Configuration/AttributedPartDiscovery.cs
davkean: Yes, and checking those paths changed the implementation. MEF v2 was retrieving import attributes twice per property, so the first instances are now reused. Export metadata was reflected twice per export, once to build metadata and again to record the defining assemblies, so assembly names are now collected during the existing metadata pass.

I also measured analogous `IsDefined` guards for type and inherited part exports, but they increased .NET Framework allocation by 1.35%, so I dropped them. Part metadata is already read in a single positive-part pass, and arbitrary export metadata cannot be skipped with one attribute-type presence check. The final base/candidate CPU, allocation, and GC results are in the PR conversation comment.

AArnott, line 70: This seems rather peculiar. For this to be an optimization I have to believe:
1. Adding `IsAttributeDefined<T>()` before `GetAttributes<T>()` does not add much overhead, and in fact does not add allocations.
2. `GetAttributes<T>()` itself is not efficiently implemented such that no attributes returns a non-allocating enumerable. That's very disappointing if true.

src/Microsoft.VisualStudio.Composition/Configuration/AttributedPartDiscoveryV1.cs
```diff
@@ -358,13 +354,21 @@ private ImportDefinitionBinding CreateImport(ParameterInfo parameter)
                 TypeRef.Get(GetImportingSiteTypeWithoutCollection(importDefinition, parameter.ParameterType), this.Resolver));
         }
 
-        private static IReadOnlyDictionary<string, object?> GetExportMetadata(MemberInfo member)
+        private static IReadOnlyDictionary<string, object?> GetExportMetadata(MemberInfo member, ISet<AssemblyName> assemblyNamesForMetadataAttributes)
```
AArnott: nit: taking an interface parameter type in a private method forces interface dispatch where it could have been a direct jump by taking a concrete type. But I suppose that's insignificant given how much other work this method does.

AArnott (approved):

src/Microsoft.VisualStudio.Composition/Configuration/AttributedPartDiscovery.cs
davkean: I was surprised too so I got Copilot to write a benchmark before I submitted. However, it sneaked in an additional allocation reduction that I didn't ask for (but was goodness), which fuddled the .NET Core numbers. I had it redo it, and shows that this by itself is slower on .NET Core, and such adding `#if` around it:

**NET Framework:** The precheck is 63% faster and saves ~50 bytes per member on.
**.NET Core:** The precheck is 300% slower.

src/Microsoft.VisualStudio.Composition/Configuration/AttributedPartDiscoveryV1.cs
davkean: Yep, good call.

src/Microsoft.VisualStudio.Composition/Configuration/AttributedPartDiscovery.cs
```diff
@@ -108,11 +113,12 @@ group export.Value by export.Key into exportsByType
 
             var exportsOnType = ImmutableList.CreateBuilder<ExportDefinition>();
             var exportsOnMembers = ImmutableDictionary.CreateBuilder<MemberRef, IReadOnlyCollection<ExportDefinition>>();
+            var assemblyNamesForMetadataAttributes = ImmutableHashSet.CreateBuilder<AssemblyName>(ByValueEquality.AssemblyName);
 
             foreach (var export in exportsByMember)
             {
                 var member = export.Key;
-                var memberExportMetadata = allExportsMetadata.AddRange(this.GetExportMetadata(member));
+                var memberExportMetadata = allExportsMetadata.AddRange(this.GetExportMetadata(member, assemblyNamesForMetadataAttributes));
```
olegtk, line 121: 💡 [suggestion] `PartDiscovery.GetAssemblyNamesFromMetadataAttributes` has no remaining callers after this change. Could you remove it so the old redundant reflection path is not mistaken for a supported helper or reused later?

olegtk (approved):

src/Microsoft.VisualStudio.Composition/Configuration/AttributedPartDiscovery.cs
davkean: Removed the now-unused helper so the redundant reflection path is no longer available for reuse.

olegtk (approved):

AArnott (approved):

